PR #62, opened 2023-09-04
feat: prepare to publish package

davidebianchi:


vite.config.ts
```diff
@@ -19,21 +19,24 @@ export default defineConfig({
...
-        'antd-5',
         'react',
         'react-dom',
       ],
-      output: {
-        globals: {
-          antd: 'antd-5',
-          react: 'React',
-          'react-dom': 'ReactDOM',
+      output: [
```
Giogia, line 28: globals configuration was used to avoid clashed with Ant versions. As an example, when a design system's component is used inside the website package, imports from ant5 will fail. This happens because Vite is not able to differentiate correctly between the two versions of Ant used (3 by website and 5 by design system). We mirrored, as a temporary solution, the package renaming convention already adopted inside website for Ant 5. We are aware that is not optimal though. 

Could you propose an alternative solution to avoid this configuration?

README.md
```diff
@@ -11,17 +11,28 @@ React graphical components designed by Mia Platform
...
 ```bash
-yarn add @mia-platform/design-system
+yarn add @mia-platform-internal/console-design-system-react
+```
+
+## Usage
+
+```jsx
+import React from 'react'
+import {Button} from '@mia-platform/design-system'
```
Danielecina, line 29:
```suggestion
import {Button} from '@mia-platform-internal/console-design-system-react'
```

vite.config.ts
davidebianchi: This lib is already a bundle with only ant5. Are you sure there is the problem?
Because in the example, in one lib there was 2 different versions of ant

Giogia (approved):

PR #85, opened 2023-09-22
feat(MNHC-56): Button

RiccardoCorona-MiaPlatform:
### Description

Button component implementation

### Issue ticket number and links

[Creazione componente Button per nuovi bottoni](https://makeitapp.atlassian.net/browse/MNHC-56)
[Design System page on Figma](https://www.figma.com/file/0ar6alIEDe8iYCb8kU7Rxd/%5BDS%5D-Console-Mia-Platform?node-id=2%3A162&mode=dev)

### TODO Checklist

- [x] Tests and stories concerning `icon` and `iconPosition`
- [ ] `button-split` implementation
- [ ] Support for Tooltip


src/components/Button/Button.tsx
```diff
@@ -0,0 +1,125 @@
...
+  type,
+}: ButtonProps): ReactElement => {
+  return (
+    <AntButton
+      className={type === 'ghost' ? ghost : undefined}
+      danger={hierarchy === 'danger'}
+      disabled={isDisabled}
+      ghost={type !== 'filled' && hierarchy !== 'neutral'}
+      href={href}
+      icon={iconPosition === 'left' && icon}
```
Danielecina, line 98: la icon position potremmo gestirla tramite css con (flex o grid) + order ("1" sx o "-1" dx)

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+import styles from './Button.module.css'
+
+const { ghost } = styles
+
+export type ButtonProps = {
+
+  /**
+   * The children nodes to be rendered with the button context
+   */
```
Danielecina, line 12: can this comment be done in a single line to reduce the file length?

src/components/Button/Button.tsx
Danielecina: I expect children handled like...
```
<Button>
     <div> ---> display flex or grid 
        <div>{icon}</div>  ---> the div with order can be 0 or -2
         <div>{children}</div>  ---> the div with sticky order 1
      </div>
</Button>
```

src/components/Button/Button.tsx
RiccardoCorona-MiaPlatform: We could, however conventionally I think this is the default way of documenting props and methods.
@Giogia what do you think about it?

Giogia: If we follow the [tsdoc](https://tsdoc.org/) comment format (/** * */) we gain some useful documentation features. There is a list of decorating tags that can be used to provide examples, verbal explanations, or links. 

Tsdoc also provides an oneliner format (/** comment */) we can use. 

However, I would rather have a common format for all the component props comments. Since it is pretty common to have multiple lines for a component property comment, I think it is better if we keep the multiline format for every property.

src/components/Button/Button.tsx
Danielecina: ok

Danielecina (approved):

src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,136 @@
+import type { Meta, StoryObj } from '@storybook/react'
+import { action } from '@storybook/addon-actions'
+
+import { Button } from './Button'
+
+const figmaUrl = 'https://www.figma.com/file/0ar6alIEDe8iYCb8kU7Rxd/%5BDS%5D-Console-Mia-Platform?node-id=2%3A162&mode=dev'
```
fredmaggiowski, line 6: is this figma public? If not why put it here if no-one can access it?

src/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+  })
+
+  test('renders danger outline button correctly', () => {
+
+    const { asFragment } = render(<Button hierarchy="danger" type="outline">{'Button'}</Button>)
+    expect(asFragment()).toMatchSnapshot()
+  })
+
+  test('renders danger ghost button correctly', () => {
+
```
fredmaggiowski, line 54: please remove redundant empty line at the beginning of each test case

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,125 @@
...
+  target,
+  type,
+}: ButtonProps): ReactElement => {
+  return (
+    <AntButton
+      className={type === 'ghost' ? ghost : undefined}
+      danger={hierarchy === 'danger'}
+      disabled={isDisabled}
+      ghost={type !== 'filled' && hierarchy !== 'neutral'}
+      href={href}
```
fredmaggiowski, line 97: when using href we should alwasy set `rel="noopener noreferrer"`

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,125 @@
...
+      onClick={onClick}
+    >
+      {children}
+      {iconPosition === 'right' && icon}
+    </AntButton>
+  )
+}
+
+Button.defaultProps = {
+  children: undefined,
```
fredmaggiowski, line 113: The default value for a prop is undefined, I'd not add all these useless defaults since they only cause noise; the actual prop that have a specific default applied are harder to grasp

src/components/Button/index.ts
```diff
@@ -0,0 +1 @@
+export { Button } from './Button'
```
fredmaggiowski, line 1: don't we have a global index that exports all of the components?

How does a user of the lib use this component?

I'd expect to write

```ts
import {Button} from 'lib'
```

and not


```ts
import {Button} from 'lib/components/Button'
```

src/components/Button/Button.module.css
```diff
@@ -0,0 +1,4 @@
+.ghost {
```
fredmaggiowski, line 1: are classes namespaced by component? or does this class apply to the whole DOM? If so I'd add a specific prefix to all of the classes we are going to define here

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,125 @@
...
+  isLoading,
+  onClick,
+  shape,
+  size,
+  target,
+  type,
+}: ButtonProps): ReactElement => {
+  return (
+    <AntButton
+      className={type === 'ghost' ? ghost : undefined}
```
fredmaggiowski, line 93: we should define enums and consts for these magic strings (`'ghost','danger','filled'....)


It would be nice to provide them to the user with something such as `Button.GhostType`

fredmaggiowski (changes requested):

src/components/Button/Button.tsx
RiccardoCorona-MiaPlatform: I can avoid specifying the default `undefined` for props only if I disable `react/require-default-props` for the entire file. Let’s consider whether to put it already as a rule of the whole design system at this point since I expect there will be other components with props not required and default values `undefined`

fredmaggiowski: I'd put the rule disabling on the whole repo tbh

src/components/Button/Button.stories.tsx
Giogia: This is how users see it without permissions, on login you are redirected again on this page

image

While without adding the figma file would appear like this:

image

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,128 @@
+/* eslint-disable react/require-default-props */
+import { ReactElement, ReactNode } from 'react'
+import { Button as AntButton } from 'antd'
+
+import { ButtonHierarchies, ButtonIconPositions, ButtonShapes, ButtonSizes, ButtonTypes } from './utils'
+import styles from './Button.module.css'
```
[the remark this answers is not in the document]
RiccardoCorona-MiaPlatform: The problem is that the CSS class name is `button-ghost` and it's not possible to extract compound names with the hyphen 🥲

src/components/Button/Button.stories.tsx
RiccardoCorona-MiaPlatform: Figma URL removed ✅

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,168 @@
...
+  isLoading,
+  onClick,
+  shape,
+  size,
+  target,
+  type,
+}: ButtonProps): ReactElement => {
+  const buttonClassNames = useMemo(() => (
+    [
+      children && size === Small && styles['button-sm'],
```
Danielecina, line 118: we felt bad here. Why not manage the class name in camelCase?

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,168 @@
...
+      children && size === Middle && styles['button-md'],
+      children && size === Large && styles['button-lg'],
+      type === Ghost && styles['button-ghost'],
+    ].filter(Boolean)
+  ), [children, size, type])
+
+  const buttonText = useMemo(() => (
+    children && <div className={styles['button-text']}>
+      {children}
+    </div>
```
Danielecina, line 128:
```suggestion
    children && (
      <div className={styles['button-text']}>
        {children}
      </div>
    )
```

oppure 

```suggestion
    children && <div className={styles['button-text']}>{children}</div>
```

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  return (
+    <AntButton
+      className={classnames(buttonClassNames)}
+      danger={hierarchy === Danger}
+      disabled={isDisabled}
+      ghost={type !== Filled && hierarchy !== Neutral}
+      href={href}
+      loading={isLoading}
+      rel={'noopener noreferrer'}
```
Danielecina, line 147: ? questo serve solo per i link esterni

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,168 @@
...
+      shape={shape === Square ? 'default' : 'circle'}
+      size={size}
+      target={target}
+      type={hierarchy === Neutral ? 'default' : 'primary'}
+      onClick={onClick}
+    >
+      {icon ? buttonWithIcon : buttonText}
```
Danielecina, line 154: usiamo solo buttonWithIcon... non abbiamo necessità di fare questa differenziazione, tanto se non passi icone è null

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * Redirect url of a link button.
+   */
+  href?: string,
+
+  /**
+   * Sets an icon component for the button.
+   */
+  icon?: ReactNode,
```
Danielecina, line 38: Qui non dovremmo mettere il tipo di Icon componente | ReactNode ?

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,168 @@
...
+   * - small: small button - 24px in height;
+   * - middle: middle button - 32px in height;
+   * - large: large button - 48px in height.
+   */
+  size?: ButtonSizes,
+
+  /**
+   * Specifies where the linked document will open when the link is clicked (default is '_blank').
+   */
+  target?: string,
```
Danielecina, line 83: mettiamo i tipi di target che ci aspettiamo. non mettiamo alcun default su black

src/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+const icon = <Icon color="white" name="PiCircleHalfTiltLight" size={16} />
+
+describe('Button Component', () => {
+  beforeEach(() => {
+    jest.resetAllMocks()
+  })
+
+  test('renders primary filled button correctly', () => {
+    const { asFragment } = render(<Button>{'Button'}</Button>)
```
Danielecina, line 21: using a container prop don't we avoid the fragment appearing in the snapshots? I would use it just in case

src/components/Button/Button.tsx
Danielecina, line 147: antd not support this props in documentation... https://ant.design/components/button#api

src/components/Button/Button.tsx
RiccardoCorona-MiaPlatform: I found online several guidelines on naming conventions for CSS classes, and all recommend using hyphens compared to camelCase. The most widespread naming convention is BEM:
https://en.bem.info/methodology/naming-convention/
https://www.freecodecamp.org/news/css-naming-conventions-that-will-save-you-hours-of-debugging-35cea737d849/
https://www.makeuseof.com/css-class-and-id-best-naming-practices/

src/components/Button/Button.tsx
RiccardoCorona-MiaPlatform: I get the error `Type 'IconType' is not assignable to type 'ReactNode'` 🥲

src/components/Button/Button.tsx
RiccardoCorona-MiaPlatform: Mmmhhh that's true @fredmaggiowski

src/components/Button/Button.tsx
RiccardoCorona-MiaPlatform: We decided not to apply this convention because it does not fit well with the use of CSS modules

src/components/Button/Button.tsx
RiccardoCorona-MiaPlatform: We could not solve the problem and we left ReactNode. In the future, it would be nice to use our interface

src/components/Button/Button.test.tsx
RiccardoCorona-MiaPlatform: It renders a `div` instead of a `DocumentFragment`, so it is a little useless

src/components/Button/Button.tsx
RiccardoCorona-MiaPlatform: We have inserted a link to document the use of `rel` (which apparently is supported by Ant if `href` is used)

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,164 @@
...
+  iconPosition,
+  isDisabled,
+  isLoading,
+  onClick,
+  shape,
+  size,
+  target,
+  type,
+}: ButtonType): ReactElement => {
+  const buttonClassNames = useMemo(() => (
```
Giogia, line 121: let's use classnames library! It's much more cleaner

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,164 @@
...
+   * - filled: button with a solid fill, typically used for primary actions;
+   * - outlined: medium-emphasis button with a stroke around the button container and no fill,
+   *   typically used for secondary actions;
+   * - ghost: transparent button with no stroke, typically used for links.
+   */
+  type?: ButtonTypes,
+}
+
+/**
+ * UI component for presenting buttons
```
Giogia, line 102: I would rather describe it as a component for performing actions on the page interacting through clicks. Otherwise the button definition is a tautology, since we use it to define it.

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,164 @@
...
+      children && size === Large && buttonLg,
+      type === Ghost && buttonGhost,
+    ].filter(Boolean)
+  ), [children, size, type])
+
+  /**
+   * rel: 'noopener noreferrer'
+   *
+   * @link https://github.com/ant-design/ant-design/blob/master/components/button/button.tsx#L259
+   */
```
Giogia, line 135: What is the purpose of this comment? I don't think it is accessible for developers on hover if we put it inside the component. I would rather add details to the href property.

Moreover, I visited the link but I found it quite difficult to grasp the meaning of its content.

src/components/Button/Button.theme.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Generates a Ant theme configuration for Button component based on a theme configuration.
+ *
+ * @link https://ant.design/components/button#design-token
+ *
+ * @param {Partial<Theme>} theme - theme configuration.
+ * @returns {Partial<ComponentsTheme>} The generated Button Ant theme configuration.
+ */
+export default ({ typography }: Partial<Theme>): ComponentsTheme['Button'] => ({
+  contentFontSizeLG: parse(typography?.bodyS?.fontSize),
```
Giogia, line 13: I would also connect contentFontSize to the action part of typography inside the theme. If we fill that part is not supporting buttons correctly, I would rather change it to have 3 different sizes. bodyS is not so correct for buttons.

src/components/Button/Button.module.css
```diff
@@ -0,0 +1,35 @@
+.button {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  gap: 4px;
+}
+
+.buttonSm {
```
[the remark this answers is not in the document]
Giogia: I feel if we use important there's another way of doing it

src/components/Button/Button.module.css
```diff
@@ -0,0 +1,35 @@
...
+
+.buttonLg {
+  display: flex;
+  align-items: center;
+  height: 48px !important;
+  padding: 16px !important;
+}
+
+.buttonGhost {
+  border-color: transparent !important;
```
Giogia, line 28: Can we use the token defaultGhostBorderColor?

src/components/Button/Button.module.css
```diff
@@ -0,0 +1,35 @@
+.button {
+  display: flex;
```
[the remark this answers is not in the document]
RiccardoCorona-MiaPlatform: If we don't apply this style the button with the icon (or the one loading) is rendered badly because buttons have by default `display: inline-block`. However, I realized that we can remove `display: flex` and `align-items: center` from all buttons styles for size (`buttonSm`, `buttonMd`, `buttonLg`) because they are already inherited from here 💪🏼

src/components/Button/Button.module.css
```diff
@@ -0,0 +1,35 @@
...
+  align-items: center;
+  justify-content: center;
+  gap: 4px;
```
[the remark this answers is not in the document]
RiccardoCorona-MiaPlatform: This was something I actually wanted to ask you yesterday 🤩 I only have a doubt about `12px` because there is no reference size in the theme padding

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * - square: square-shaped button with rounded corners;
+   * - circle: fully rounded button, typically used for buttons containing only icon.
+   */
+  shape?: ButtonShapes,
+
+  /**
+   * Defines the button size. Either:
+   *
+   * - small: small button - 24px in height;
```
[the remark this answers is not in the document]
RiccardoCorona-MiaPlatform: Oooh right, I didn't think of that. I will specify then that those values refer to the default case

src/components/Button/Button.tsx
RiccardoCorona-MiaPlatform: I used it below for the `join` operation but I didn't know it could also filter away falsy values. I'll replace it right away 👍🏼

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+const { button, buttonSm, buttonMd, buttonLg, buttonGhost, buttonText } = styles
+
+const { Primary, Neutral, Danger } = ButtonHierarchies
+const { Left, Right } = ButtonIconPositions
+const { Square } = ButtonShapes
+const { Small, Middle, Large } = ButtonSizes
+const { Filled, Ghost } = ButtonTypes
+
+export type ButtonType = {
```
Giogia, line 16: better ButtonProps, we need to change it on Icon as well

src/components/Button/Button.tsx
RiccardoCorona-MiaPlatform: We chose to put that link because it represents the only reference "documentation" for the fact that we put `rel` between the props of the Ant component. The Ant button does not document `rel` among the possible props, but in fact, accepts it and makes the spread within an HTML component of type `<a>`, as seen in the link.
I can move the link to the `href` section and try to provide this explanation there as well, although it seems slightly out of scope

src/components/Button/Button.theme.ts
RiccardoCorona-MiaPlatform: I agree, but currently `action` has `"fontSize": "16"`, while our buttons always have 14px. Should we change the value in the theme file?

src/components/Button/Button.module.css
RiccardoCorona-MiaPlatform: Unfortunately not, because that token affects the behavior of `ghost` only for what Ant defines as the default button (what would be our neutral). So for ghost primary and danger, the style would not be applied

src/components/Button/Button.tsx
Danielecina: We add that ref works because in the case of prop href and uses an html <a> tag

Giogia: I think a person without context can have some difficulties understanding that. I would try to add some explanation details to make it more clear. 

However, I don't think this comment is necessary at all. As long as the Ant Button component type does not raise errors on such a property, I think we can conclude it is actually supported. 
In addition, the rel attribute is also very harmless. In general, applying a rel attribute on HTML tags that do not accept it represents more of a bad practice than a blocking error that can break the dom.

src/components/Button/Button.module.css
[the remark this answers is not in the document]
Giogia: Whenever we feel like the theme configuration is not expressive enough for our needs, we can have a discussion with the design theme and enhance the theme configuration

src/components/Button/Button.theme.ts
Giogia: I would actually add a actionLg property to the theme for the contentFontSizeLG token, and use action for the contentFontSize token.

src/components/Button/Button.module.css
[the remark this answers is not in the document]
Giogia: ok, fair enough

src/components/Button/Button.module.css
Giogia: I see. ok for now then :+1:

src/components/Button/Button.tsx
RiccardoCorona-MiaPlatform: I removed the comment and added in the documentation on `href` the following sentence: `For security reasons, the attribute "rel: 'noopener noreferrer'" is always specified.`
I think it's the best compromise 🤝🏼🤝🏼

Giogia (approved):

Danielecina (approved):

src/components/Button/Button.tsx
fredmaggiowski: It works :)

PR #102, opened 2023-09-28
feat(PDS-58): Headings

RiccardoCorona-MiaPlatform:
### Description

Headings components implementation

[Componente Headings](https://makeitapp.atlassian.net/browse/PDS-58)

### TODO Checklist

- [x] Support for Ellipsis


src/components/Typography/HX/HX.tsx
```diff
@@ -0,0 +1,62 @@
+import { ReactElement, ReactNode } from 'react'
+import { Typography as AntTypography } from 'antd'
+
+import { CopyConfig, EllipsisConfig } from '../Typography.types'
+
+const { Title: AntTitle } = AntTypography
+
+export type HXProps = {
+
```
fredmaggiowski, line 9: We decided not to uese these new lines

src/components/Typography/HX/HX.tsx
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Displays ellipsis when text overflows. By default, the text will be truncated if it
+   * exceeds the length of 2 rows, showing all informational content in a tooltip.
+   * The attribute is further customizable.
+   */
+  ellipsis?: boolean | EllipsisConfig,
+}
+
+type HXLevel = {
+
```
fredmaggiowski, line 30: Same

src/components/Typography/Typography.types.ts
```diff
@@ -0,0 +1,76 @@
+import { TooltipProps } from 'antd'
+
+/**
+ * @link https://ant.design/components/typography#copyable
+ */
+export type CopyConfig = {
+
```
fredmaggiowski, line 7: Same new line

src/components/Button/__snapshots__/Button.test.tsx.snap
```diff
@@ -85,7 +85,7 @@ exports[`Button Component renders button with icon right correctly 1`] = `
 exports[`Button Component renders circle button correctly 1`] = `
 <DocumentFragment>
   <button
-    class="ant-btn css-dev-only-do-not-override-1vr7spz ant-btn-circle ant-btn-primary button"
+    class="ant-btn css-dev-only-do-not-override-1vr7spz ant-btn-circle ant-btn-primary button buttonMdIconOnly"
```
fredmaggiowski, line 88: Why has this class appeared? Also why in camelCase instead of the typical kebab-case?

fredmaggiowski (changes requested):

src/components/Button/__snapshots__/Button.test.tsx.snap
RiccardoCorona-MiaPlatform: It is a remnant of the PR on Button, where I had forgotten to update the snapshots (we should configure the pipeline so that it blocks the merge operation if the pipeline fails). The use of camelCase is due to the fact that it is a class set by us via CSS, and with the use of CSS modules it is more convenient (there have already been discussions about it in the PR on Button)

src/components/Typography/HX/HX.tsx
RiccardoCorona-MiaPlatform: We decided not to have a new line between a function declaration and its implementation, but here we are documenting the properties of the component and I think the new line slightly improves readability since it will always be followed by a comment

src/components/Icon/Icon.tsx
```diff
@@ -50,7 +50,7 @@ export type IconProps = {
 }
 
 /**
- * UI component for displaying different icon packs (Ant, Feather, Phosphor) and custom SVGs
+ * UI component to display different icon packs (Ant, Feather, Phosphor) and custom SVGs
```
Giogia, line 53: In this sentence, the absence of "for" makes it more concise and suggests that the UI component is capable of displaying different icon packs and custom SVGs, but it doesn't necessarily imply that it is designed solely for that purpose. It's a statement of functionality without specifying the primary purpose.

In practical terms, both sentences convey a similar idea, but the first one emphasizes the specific purpose or intent of the UI component, while the second one simply states what it can do. The choice between them depends on the context and how you want to frame the information.

Giogia: Also, this comment is a little bit out of scope. I would try to keep modifications to other components only if strictly necessary.

src/components/Typography/HX/HX.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { loremIpsum } from '../../../utils/loremIpsum'
+
+const customEllipsis = {
+  rows: 1,
+  suffix: ' END',
+  expandable: true,
+  symbol: 'Show more',
+}
+
+const customCopyable = {
```
Giogia, line 15: Can we move these resources to a Typography.mocks.ts file? in this way, we can also reuse them in tests.

src/components/Typography/HX/HX.test.tsx
```diff
@@ -0,0 +1,74 @@
+import { render } from '@testing-library/react'
+
+import { Icon } from '../../Icon'
+import { Typography } from '..'
+import { loremIpsum } from '../../../utils/loremIpsum'
+
+const { H1, H2, H3, H4 } = Typography
+
+describe('HX Component', () => {
```
Giogia, line 9: We need to add some tests in which we try to get the specific element from the screen in an easy way. ( e.g. screen.getByRole('something', {name: 'something' }) and screen.getyText('something')  )

Giogia: I would also try interactions with tooltip and copyable icon.

src/components/Typography/Typography.theme.ts
```diff
@@ -0,0 +1,27 @@
+import { ComponentsTheme, parse } from '../ThemeProvider/Ant'
+import Theme from '../../themes/schema'
+
+const getRelativeLineHeight = (lineHeight?: number, fontSize?: number): number | undefined => {
```
Giogia, line 4: Can we add tsdoc to this function?

src/utils/loremIpsum.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Utils file to export texts of various lengths to be used in stories and tests.
+ */
+
```
Giogia, line 4:
```suggestion
```

src/components/Typography/HX/HX.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+type Story = StoryObj<typeof meta>
+
+export const HXComponent: Story = {
+  args: { ...meta.args, level: 1 },
+}
+
+export const All = {
+  argTypes: {
+    children: { table: { disable: true } },
+    level: { table: { disable: true } },
```
Giogia, line 53: I would remove the level declaration from every story and add it to meta argTypes. Then I would enable it specifically for the HXComponent story

src/components/Typography/HX/HX.tsx
```diff
@@ -0,0 +1,62 @@
...
+export type HXProps = {
+
+  /**
+   * The children nodes to be rendered within the typography context.
+   */
+  children?: ReactNode,
+
+  /**
+   * Whether the text is copyable. If set to true, a copy icon is shown to the right of the text.
+   * The attribute is further customizable.
```
Giogia, line 17: Can we document better? I would add the entire data model or at least a link to its definition.

src/components/Typography/HX/HX.tsx
```diff
@@ -0,0 +1,62 @@
...
+   */
+  copyable?: boolean | CopyConfig,
+
+  /**
+   * Displays ellipsis when text overflows. By default, the text will be truncated if it
+   * exceeds the length of 2 rows, showing all informational content in a tooltip.
+   * The attribute is further customizable.
```
Giogia, line 24: I would also add details to this

src/components/Typography/HX/HX.tsx
```diff
@@ -0,0 +1,62 @@
...
+type HXLevel = {
+
+  /**
+   * The H tag to be rendered. Match with H1, H2, H3, H4.
+   */
+  level: 1 | 2 | 3 | 4,
+}
+
+/**
+ * UI component to display headers (H1, H2, H3, H4).
```
Giogia, line 38:
```suggestion
 * UI component for displaying headers (H1, H2, H3, H4).
```

As I explained before, the word "to" declare the ability of the component rather than its core design to complete a specific functionality.

src/components/Typography/HX/HX.tsx
```diff
@@ -0,0 +1,62 @@
...
+      ellipsis={ellipsis ?? { rows: 2, tooltip: { children } }}
+      level={level}
+    >
+      {children}
+    </AntTitle>
+  )
+}
+
+HX.defaultProps = {
+  copyable: false,
```
Giogia, line 61: ellipsis is undefined by default. Is this correct for its typings? I think we should re-add the eslint rule for default prop declaration. In this way, we don't need to care about these edge cases manually every time.

src/components/Typography/Typography.types.ts
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * The text to copy.
+   */
+  text?: string;
+
+  /**
+   * Function called when the text is copied.
+   */
+  onCopy?: (event?: React.MouseEvent<HTMLDivElement>) => void;
```
Giogia, line 16: Have you tried to valorize this function? This type usually raises an error if a function without parameters is passed.

src/components/Typography/Typography.types.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Custom copy icon(s): [copyIcon, copiedIcon].
```
Giogia, line 19: it is not clear if both a single icon and an array are correct values for this property.

src/components/Typography/Typography.types.ts
```diff
@@ -0,0 +1,76 @@
...
+   */
+  icon?: React.ReactNode;
+
+  /**
+   * Custom tooltip text(s): [copyText, copiedText]. If the attribute is set to false, no tooltip will be shown.
```
Giogia, line 24: It is not clear if both a single icon and an array of icons can be passed, I would try to specify it.

src/components/Typography/Typography.types.ts
```diff
@@ -0,0 +1,76 @@
...
+   */
+  tooltips?: boolean | React.ReactNode;
+
+  /**
+   * The Mime Type of the text.
```
Giogia, line 29: Can We add this link here ? https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

src/themes/files/light/theme-generator.json
```diff
@@ -443,9 +443,9 @@
     "h1": {
       "value": {
         "fontFamily": "{fontFamilies.main}",
+        "fontWeight": "{fontWeights.semi-bold}",
         "lineHeight": "{lineHeights.3}",
```
Giogia, line 447: What is the purpose of this?

src/themes/files/light/theme.json
```diff
@@ -148,9 +148,9 @@
   "typography": {
     "h1": {
       "fontFamily": "Inter",
+      "fontWeight": "600",
```
Giogia, line 151: what is the purpose of this?

src/components/Button/__snapshots__/Button.test.tsx.snap
Giogia, line 88: This is out of scope. It should be fixed in a specific merge request

Giogia (changes requested):

src/utils/loremIpsum.ts
Giogia, line 1: I don't think this file suits the utils directory very well. Usually what I intend for utils is plain helper functions. I would rather add a src/mocks/ directory for this type of content. 

Another option is to add this mock directly on Typography component. Since its scope is handling texts, other component might import it from them.
import {loremIpsum} from 'Typography.mocks'

Giogia (changes requested):

src/themes/files/light/theme.json
```diff
@@ -148,9 +148,9 @@
...
       "lineHeight": "32",
-      "fontSize": "24",
-      "fontWeight": "600"
+      "fontSize": "24"
```
Danielecina, line 153: these values are intended as pixels?

src/utils/loremIpsum.ts
```diff
@@ -0,0 +1,13 @@
...
+export const loremIpsum = `Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque vel dolor ac sem congue ornare rhoncus quis tortor.
```
Danielecina, line 5: we can use a dev dependencies calls https://www.npmjs.com/package/lorem-ipsum to generate dynamically words or sentence

Danielecina: a mock should be used to make it work. 
I suggest creating a testsUtils folder for this utility

src/utils/loremIpsum.ts
RiccardoCorona-MiaPlatform: Isn't it better to maintain the new line in this case since the comment refers to the entire file and not to the const below?

src/components/Button/__snapshots__/Button.test.tsx.snap
RiccardoCorona-MiaPlatform: It is a remnant of the PR on Button, where I had forgotten to update the snapshots (we should configure the pipeline so that it blocks the merge operation if it fails). To me, it seems a little superfluous to open an ad-hoc PR

src/themes/files/light/theme-generator.json
RiccardoCorona-MiaPlatform: I simply rearranged the attribute declarations, since for all subsequent typography elements the order was the same (`fontFamily`, `fontWeight`, `lineHeight`, `fontSize` respectively). This reordering is also obviously reflected in the generated theme

Giogia: We cannot change the order of this structure since it is automatically generated by a Figma token plugin. We need to make sure with the design team whenever we change something in the theme.

Giogia (changes requested):

src/components/Typography/HX/HX.stories.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { customCopyable, customEllipsis } from '../Typography.mocks'
+import { HX } from './HX'
+import { Typography } from '..'
+import { loremIpsum } from '../../../utils/loremIpsum'
+
+const meta = {
+  title: 'Components/Typography/HX',
+  component: HX,
+  args: {
+    ...HX.defaultProps,
```
Giogia, line 12: it is a bit hard to work with copyable and ellipsis objects in the docs panel. Can we set them to boolean? It is true that we lose the customization, but we also have examples in later stories

src/themes/files/light/theme-generator.json
RiccardoCorona-MiaPlatform: Alright, I would then communicate to the design team even this small change, together with the modifications to spacers and font size that we applied in the PR on Button

src/components/Typography/HX/HX.tsx
RiccardoCorona-MiaPlatform: I solved it another way. The prop `ellipsis` now defaults to `true` and, in that case, our "standard" ellipsis is applied. The user can also pass a custom ellipsis that will be applied instead of the standard, or `false` to disable the ellipsis

src/components/Typography/Typography.types.ts
RiccardoCorona-MiaPlatform: I replaced it with our usual `() => void`

src/themes/files/light/theme.json
RiccardoCorona-MiaPlatform: The theme is automatically generated by the `theme-generator.json` file 😬 we should update it properly by common agreement with the design team

src/utils/loremIpsum.ts
RiccardoCorona-MiaPlatform: I took a look at the library and it looks like a nice tool. However, it may be a bit too much for the use I actually think we'll make of it in the design system (we could use it for website though 🤔). For the moment, I imported the const inside Typography.mocks.ts, with a comment saying to move it in case it should be used for stories or tests outside Typography

src/utils/loremIpsum.ts
RiccardoCorona-MiaPlatform: File removed

Giogia:
We need to add Apache licenses on each file.

src/components/Typography/HX/HX.tsx
RiccardoCorona-MiaPlatform: Same as above

src/components/Typography/Typography.types.ts
RiccardoCorona-MiaPlatform: Same as above

Giogia (approved):

fredmaggiowski (approved):

PR #103, opened 2023-09-29
feat(PDS-63): segmented control

Giogia:
### Description

Segmented component implementation

 #### Internal Analysis

https://makeitapp.atlassian.net/wiki/spaces/MP/pages/4182016050/Componente+Segmented

[Componente Segmented](https://makeitapp.atlassian.net/wiki/spaces/MP/pages/4182016050/Componente+Segmented)

### Checklist

- [x] added the link to the Jira task
- [x] tests are included


src/components/Segmented/Segmented.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+import { Option, OptionsAlignments } from './Segmented.types'
+import { getOptionKey, isDisabledOption, isString, isVerticalOption, resolveKey } from './Segmented.utils'
+import styles from './Segmented.module.css'
+
+const { Horizontal, Vertical } = OptionsAlignments
+const { segmented, segmentedOption, disabled, selected, vertical } = styles
+
+export type SegmentedProps = {
+
```
fredmaggiowski, line 12: Didn't we decide to remove these new lines?

fredmaggiowski (commented):
Shouldn't the component be exposed on the main index?

src/components/Segmented/Segmented.tsx
Giogia: We already removed the padded blocks rule. This space is different, it is enforced because it is a comment (eslint rule lines-around-comment) which comes with the recommended default configuration. If we want to remove this space we need to change that rule. Why does this space represent a problem to you?

src/components/Segmented/Segmented.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+          padding: 8,
+        }}
+      >
+        <Story />
+      </div>
+    </div>
+  )],
+}
+
+
```
RiccardoCorona-MiaPlatform, line 100:
```suggestion
```

src/components/Segmented/Segmented.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+    test('calls onChange correctly', () => {
+      render(<Segmented {...props} />)
+
+      fireEvent.click(screen.getByRole('listitem', { name: clickedOption }))
+
+      expect(onChange).toBeCalledTimes(1)
+      expect(onChange).toBeCalledWith(clickedOption)
+    })
+  })
+})
```
RiccardoCorona-MiaPlatform, line 199:
```suggestion
})

```

src/components/Segmented/Segmented.tsx
```diff
@@ -0,0 +1,143 @@
...
+   * - object: {
+   *
+   *     icon?: ReactNode, The icon of the option
+   *
+   *     label: ReactNode, The selectable text of the option
+   *
+   *     isDisabled?: boolean, Whether the option is clickable
+   *
+   *     key: string, The value associated with the option
+   *
```
RiccardoCorona-MiaPlatform, line 45:
```suggestion
   *     icon?: ReactNode - The icon of the option
   *     label: ReactNode - The selectable text of the option
   *     isDisabled?: boolean - Whether the option is clickable
   *     key: string - The value associated with the option
```

src/components/Segmented/Segmented.test.tsx
RiccardoCorona-MiaPlatform: Can we set up an ESLint rule to automatically place one line at the end of the file? 🤔

src/components/Segmented/Segmented.utils.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/**
+ * Retrieves a key from the provided value or options object.
+ *
+ * @param {Option[]} options - An object containing options for key retrieval.
+ * @param {string | number} value - The key to retrieve, or an options position containing the key.
+ *
+ * - string: retrieves the option with value as key
+ * - number: retrieves the option from options in the value position
```
RiccardoCorona-MiaPlatform, line 62:
```suggestion
 * @param {Option[]} options - An array containing options for key retrieval.
 * @param {string | number} value - The key to retrieve, or an options position containing the key.
 *
 * - string: retrieves the option with value as key
 * - number: retrieves the option from the options array from the position specified by the value
```

src/components/Segmented/Segmented.tsx
Danielecina: I agree with Fede, let's remove these lines. They don't help that much with readability and I would start thinking about putting the types in a separate file, what do you think?

src/components/Segmented/Segmented.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const [selectedValue, setSelectedValue] = useState(resolveKey(options, defaultValue))
+
+  const handleOptionClick = (option: Option): void => {
+    if (!isDisabledOption(option, isDisabled)) {
+      setSelectedValue(getOptionKey(option))
+      if (onChange) {
+        onChange(option)
+      }
+    }
+  }
```
Danielecina, line 89: why don't we use useCallback here? the only dependency is onChange.

src/components/Segmented/Segmented.tsx
```diff
@@ -0,0 +1,143 @@
...
+  return (
+    <ul
+      className={classnames([
+        segmented,
+        isDisabled && disabled,
+        optionsAlignment === Vertical && vertical,
+      ])}
+    >
+      {
+        options.map((option) => {
```
Danielecina, line 100: if we move this map up to one line, we reduce the indentation by several lines

src/components/Segmented/Segmented.tsx
```diff
@@ -0,0 +1,143 @@
...
+              {
+                isString(option)
+                  ? option
+                  : (
+                    <>
+                      {option.icon}
+                      {option.label}
+                    </>
+                  )
+              }
```
Danielecina, line 131: these ternaries made like this don't add up to me. We use the default of eslint. https://eslint.org/docs/latest/rules/multiline-ternary

```suggestion
              {isString(option) ? option : <>{option.icon}{option.label}<>}
```

src/components/Segmented/Segmented.utils.ts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Checks if an option is disabled.
+ *
+ * @param {Option} option - The option to check.
+ * @param {boolean} isDisabled - Specifies whether the whole segmented is disabled.
+ *
+ * @returns {boolean} `true` if the option is disabled, `false` otherwise.
+ */
+export function isDisabledOption(option: Option, isDisabled = false): boolean {
+  return isDisabled ?? (
```
Danielecina, line 24: The nullish operator returns the right-hand value if the left-hand value is null or undefined. 
In this case, the default value is false. 
So this function will return false if the value is undefined. 
How it works? Am I missing something?

src/components/Segmented/Segmented.utils.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @param {Option} option - The option to check.
+ * @param {OptionsAlignments} optionsAlignment - Specifies the alignment of segmented options.
+ *
+ * @returns {boolean} `true` if the option should be aligned vertically, `false` otherwise.
+ */
+export function isVerticalOption(option: Option, optionsAlignment = Horizontal): boolean {
+  return (
+    !isString(option) && optionsAlignment === Vertical
+  )
```
Danielecina, line 40:
```suggestion
  return !isString(option) && optionsAlignment === Vertical
```

src/components/Segmented/Segmented.module.css
```diff
@@ -0,0 +1,49 @@
...
+  background-color: color-mix(in srgb, var(--palette-common-grey-300, #e6e6e6), transparent 50%);
+}
+
+.segmentedOption {
+  flex: 1;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  text-align: center;
+  gap: calc(var(--spacing-gap-xs, 4) * 1px);
```
Danielecina, line 22: Why do we use Calc here?

src/components/Segmented/Segmented.module.css
```diff
@@ -0,0 +1,49 @@
...
+  border-radius: calc(var(--shape-border-radius-md, 4) * 1px);
+  cursor: pointer;
+  transition: background-color 0.25s ease-in-out;
+}
+
+.segmentedOption:hover {
+  background-color: color-mix(in srgb, var(--palette-common-grey-600, #aaaaaa), transparent 75%);
+}
+
+.segmentedOption.selected {
```
Danielecina, line 34: Why do we use a double class here? it should just be enough ".selected" and ".vertical"

src/components/Segmented/Segmented.tsx
Giogia: Regarding the name of the component instead, it is a design choice. Nothing to do with the work expressed in this merge request. It is also a very typical name for this kind of interaction.

src/components/Segmented/Segmented.tsx
Giogia: This also comes witht a slightly less readable Storybook.

It would go from this
image

to this
image

to me it is not worth it.

src/components/Segmented/Segmented.module.css
Giogia: because it does not work without specifying pixels

src/components/Segmented/Segmented.tsx
fredmaggiowski: I just found out that antd has the same kind of component with the same name but it's the first time I see a button group with that name tbh, maybe "SegmentedButton" would be a clearer name?

fredmaggiowski: To better address the name issue I've done a little bit of research:

* antd calls this component "segmented" https://ant.design/components/segmented 
* many other calls it segment controls:
   * uber: https://base.uber.com/6d2425e9f/p/05421e-segmented-control/b/46994b
   * apple: https://developer.apple.com/design/human-interface-guidelines/segmented-controls
   * gitlab: https://design.gitlab.com/components/segmented-control
   * microsoft: https://fluent2.microsoft.design/components/ios/segmentedcontrol/usage

Danielecina: The most used name is https://component.gallery/components/segmented-control/ SegmentedControls. Can it be called like this?

src/components/Segmented/Segmented.tsx
Danielecina: Because code with more indentation is less readable, and having the word option next to a parenthesis doesn't make it any less readable

src/components/Segmented/Segmented.tsx
Danielecina: OK, you cannot use the "one-line" proposal, but I use the interrogative and exclamatory sentences at the end of the line and not at the start. Let's align ourselves with the default

src/components/Segmented/Segmented.module.css
Danielecina: OK, but I suggest changing the variable directly, otherwise we have to do this thing every time

src/components/Segmented/Segmented.test.tsx
RiccardoCorona-MiaPlatform: I think it's a common best practice to have one and only empty line at the end of the file (GitLab itself advises you when it's missing, saying `\ No newline at end of file`)

src/components/Segmented/Segmented.tsx
RiccardoCorona-MiaPlatform: I didn’t think the visualization in the Storybook would be so drastically altered. Fair enough 👍🏼

src/components/Segmented/Segmented.utils.ts
Giogia: I changed it, now the function accept a general boolean isDisabled without default prop

src/components/Segmented/Segmented.test.tsx
Giogia: Maybe I misunderstood the first time, it should be already the way you are saying. I thought you meant an empty line and a new line

src/components/Segmented/Segmented.tsx
```diff
@@ -0,0 +1,141 @@
...
+   *
+   *     icon?: ReactNode, The icon of the option
+   *
+   *     label: ReactNode, The selectable text of the option
+   *
+   *     isDisabled?: boolean, Whether the option is clickable
+   *
+   *     key: string, The value associated with the option
+   *
+   *   }
```
RiccardoCorona-MiaPlatform, line 46:
```suggestion
   * - `object: {`
   *
   *     `icon?: ReactNode` - The icon of the option
   *
   *     `label: ReactNode` - The selectable text of the option
   *
   *     `isDisabled?: boolean` - Whether the option is clickable
   *
   *     `key: string` - The value associated with the option
   *
   *   `}`
```

RiccardoCorona-MiaPlatform: If you don't like it go ahead and resolve the comment and write me so I can adjust it in my PR as well 🤙🏼

src/components/Segmented/Segmented.test.tsx
RiccardoCorona-MiaPlatform: You are right, if I try to delete the last line I get the error `Newline required at end of file but not found.` and saving it is automatically added. Can it be GitHub that doesn't let me see it then? 🤔

Giogia: Yes, I think Github is purposely removing the last newline of each file in its UI.

src/components/Segmented/Segmented.tsx
Giogia: Nice hint! What do you think about wrapping only types? sub fields are still emphasized correctly but we don't lose the classic format of Storybook.

RiccardoCorona-MiaPlatform: When you talk about types do you mean only `string`, `boolean`, and so on? That sounds okay to me, and would you also include the `object` that wraps everything?

src/components/Segmented/Segmented.module.css
Giogia: We discussed this topic with the design theme. We could actually use measures with units (adding px) on the theme design tool, but this modification could have some drawbacks in setting the Ant theme, whose tokens are usually numbers. We would have at that point to remove the unit measure from the string. 

I think we need to face this problem in a specific merge request, considering both the pros and cons of this approach. For the moment, I believe it is safer to stick with numbers and calc.

src/components/Segmented/Segmented.tsx
Giogia: Discussing with the design team, the final decision for the component name is *Segmented Control*

Giogia:
> Shouldn't the component be exposed on the main index?
> 
> What is behind the choice of the Segmented name? I struggled in understanding what the final component wouldnlook like and I thought it would like like typical breadcrumbs while instead it's more like a button group, what should the name segmented recall to the user? Can we find a more explicit name?

I think both the name of the component and the export issues have been addressed

src/components/Segmented/Segmented.stories.tsx
Giogia: removed :+1:

src/components/Segmented/Segmented.utils.ts
Giogia: I rephrased the doc to make it clearer.

src/components/Segmented/Segmented.tsx
Giogia: ok it is fine to remove the space

src/components/Segmented/Segmented.module.css
Giogia: I can only put selected, if I include also vertical the whole shape of the control goes vertical, which is not the expected behaviour.

Giogia: I tried but in both cases, the style did not get the expected behavior. 

For vertical, the whole shape of the control aligns vertically.
For selected, the background color collides when on hover, keeping the hover color rather than the selected color

src/components/Segmented/Segmented.tsx
Giogia: I followed your suggestion and now it is like this:

image

I think it clear enough now

src/components/Segmented/Segmented.tsx
Giogia: added callback

fredmaggiowski:
> > Shouldn't the component be exposed on the main index?
> > What is behind the choice of the Segmented name? I struggled in understanding what the final component wouldnlook like and I thought it would like like typical breadcrumbs while instead it's more like a button group, what should the name segmented recall to the user? Can we find a more explicit name?
> 
> I think both the name of the component and the export issues have been addressed

mmmh @Giogia I don't see changes in the [main index file](https://github.com/mia-platform/design-system/blob/main/src/index.ts) 🤔

src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+ /**
+   * An array of segmented options to be displayed. Either one of the following:
+   *
+   * - `string`
+   * - `object`:
+   *    - icon?: The icon of the option `ReactNode`
+   *    - label: The selectable text of the option `ReactNode`
+   *    - isDisabled?: Whether the option is clickable `boolean`
+   *    - key: The value associated with the option `string`
```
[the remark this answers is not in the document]
RiccardoCorona-MiaPlatform: Small trick to make type statements go to the head without taking up 4 lines 🤭

Screenshot from 2023-10-04 15-24-08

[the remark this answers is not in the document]
Giogia: awesome, added!

Giogia:
> > > Shouldn't the component be exposed on the main index?
> > > What is behind the choice of the Segmented name? I struggled in understanding what the final component wouldnlook like and I thought it would like like typical breadcrumbs while instead it's more like a button group, what should the name segmented recall to the user? Can we find a more explicit name?
> > 
> > 
> > I think both the name of the component and the export issues have been addressed
> 
> mmmh @Giogia I don't see changes in the [main index file](https://github.com/mia-platform/design-system/blob/main/src/index.ts) 🤔

My bad, now should be done

fredmaggiowski (approved):
LGTM

Danielecina (approved):

RiccardoCorona-MiaPlatform (approved):

PR #144, opened 2023-10-20
fix: segmented control disabled and vertical items

Giogia:
##### SegmentedControl

Segmented control can now correctly singularly disable options, it also renders the primary hierarchy correctly (vertical and disabled options were affected).

### Checklist

- [x] tests are included

### DIFF
--- a/src/components/SegmentedControl/SegmentedControl.utils.ts
+++ b/src/components/SegmentedControl/SegmentedControl.utils.ts
@@ -39,7 +39,7 @@ export function isString(element: Option | number): element is string {
  * @returns {boolean} `true` if the option is disabled, `false` otherwise.
  */
 export function isDisabledOption(option: Option, isDisabled: boolean): boolean {
-  return isDisabled || (!isString(option) && option?.isDisabled)
+  return isDisabled || (!isString(option) && option?.isDisabled) || false
 }
 
 /**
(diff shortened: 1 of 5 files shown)


fredmaggiowski (approved):
LGTM

RiccardoCorona-MiaPlatform (dismissed):

src/components/SegmentedControl/SegmentedControl.utils.ts
```diff
@@ -39,7 +39,7 @@ export function isString(element: Option | number): element is string {
  * @returns {boolean} `true` if the option is disabled, `false` otherwise.
  */
 export function isDisabledOption(option: Option, isDisabled: boolean): boolean {
-  return isDisabled ?? (!isString(option) && option?.isDisabled)
+  return (!isString(option) && option?.isDisabled) || isDisabled
```
davidebianchi, line 42: Why those change? It is required?
I think that the previous check is correct, if the component is disabled globally, so it should exit for isDisabled. Otherwise, it checks inside the option. 
What do you think about? Maybe it works only changing the nullish operator with the or?

Giogia: yes, we can favor isDisabled first

Giogia, line 42:
```suggestion
  return isDisabled || (!isString(option) && option?.isDisabled)
```

RiccardoCorona-MiaPlatform (approved):

davidebianchi (approved):

PR #152, opened 2023-10-24
fix: segmented control pass event to onChange, fix button tests

Giogia:
##### [Button]

Added mouse event type to onClick, fixed tests 

##### [Segmented Control]

Added mouse event parameter to onChange function

### Checklist

- [x] tests are included

### DIFF
--- a/src/components/Button/Button.test.tsx
+++ b/src/components/Button/Button.test.tsx
@@ -136,5 +136,12 @@ describe('Button Component', () => {
     const button = screen.getByRole('button', { name: /Button/i })
     fireEvent.click(button)
     expect(onClick).toHaveBeenCalledTimes(1)
+    expect(onClick).toBeCalledWith(expect.objectContaining({
+      altKey: expect.any(Boolean),
+      ctrlKey: expect.any(Boolean),
+      metaKey: expect.any(Boolean),
+      shiftKey: expect.any(Boolean),
+      target: expect.any(HTMLElement),
+    }))
   })
 })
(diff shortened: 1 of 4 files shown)


RiccardoCorona-MiaPlatform (dismissed):

src/components/Button/Button.test.tsx
```diff
@@ -136,5 +136,12 @@ describe('Button Component', () => {
     const button = screen.getByRole('button', { name: /Button/i })
     fireEvent.click(button)
     expect(onClick).toHaveBeenCalledTimes(1)
+    expect(onClick).toBeCalledWith(expect.objectContaining({
```
davidebianchi, line 139: It is needed to add this test? We should expect a MouseEvent maybe?

src/components/SegmentedControl/SegmentedControl.test.tsx
```diff
@@ -136,7 +136,13 @@ describe('Segmented Control Component', () => {
       fireEvent.click(screen.getByRole('listitem', { name: clickedOption.key }))
 
       expect(onChange).toBeCalledTimes(1)
-      expect(onChange).toBeCalledWith(clickedOption)
+      expect(onChange).toBeCalledWith(clickedOption, expect.objectContaining({
```
davidebianchi, line 139: Also here

davidebianchi (changes requested):

Danielecina (dismissed):

src/components/Button/Button.test.tsx
Giogia: I tried using expect.any(MouseEvent) but it fails. 
This to me was the closest way to test what we need from a MouseEvent. 

Would you have any recommendations on how to improve it?

Giogia: we swtiched to expect(onChange).toBeCalledWith(expect.objectContaining({ ...MouseEvent })), still not the best but should be closer to what we expect

src/components/SegmentedControl/SegmentedControl.test.tsx
Giogia: Same approach adopted in the other comment

PR #247, opened 2024-01-05
Card

RiccardoCorona-MiaPlatform:
### Description

Card component implementation

### Issue ticket number and links

[Creazione componente Card](https://makeitapp.atlassian.net/browse/PDS-37)
[Design System page on Figma](https://www.figma.com/file/0ar6alIEDe8iYCb8kU7Rxd/%5BDS%5D-Console-Mia-Platform?node-id=1582%3A10787&mode=dev)

### TODO Checklist

- [ ] Support for Tabs
- [ ] Support for Tags
- [ ] Support for Switch

### DIFF
--- a/src/components/Card/Card.stories.tsx
+++ b/src/components/Card/Card.stories.tsx
@@ -1,0 +1,129 @@
+/**
+ * Copyright 2023 Mia srl
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { Meta, StoryObj } from '@storybook/react'
+
+import { Hierarchy, Shape, Type } from '../Button/Button.types'
+import { columns, data, rowKey } from '../Table/Table.mocks'
+import { Button } from '../Button'
+import { Card } from '.'
+import { Icon } from '../Icon'
+import { Table } from '../Table'
+
+const { Danger } = Hierarchy
+const { Circle } = Shape
+const { Outlined } = Type
+
+const actionIcon = <Icon color="#b91200" name="PiTrash" size={16} />
+const actionButton = (
+  <Button
+    hierarchy={Danger}
+    icon={actionIcon}
+    shape={Circle}
+    type={Outlined}
+  />
+)
+
+const meta = {
+  component: Card,
+  args: {
+    ...Card.defaultProps,
+    children: 'Card Content',
+    docLink: 'https://mia-platform.eu',
+    extra: <Button>{'Button'}</Button>,
+    subtitle: 'Card Subtitle',
+    title: 'Card Title',
+  },
+} satisfies Meta<typeof Card>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    ...meta.args,
+  },
+}
+
+export const Title: Story = {
+  args: {
+    children: undefined,
+    docLink: undefined,
+    extra: undefined,
+    subtitle: undefined,
+  },
+}
+
+export const TitleWithDocLink: Story = {
+  args: {
+    children: undefined,
+    extra: undefined,
+    subtitle: undefined,
+  },
+}
+
+export const HeaderWithSubtitle: Story = {
+  args: {
+    children: undefined,
+    extra: undefined,
+  },
+}
+
+export const HeaderWithAction: Story = {
+  args: {
+    children: undefined,
+    subtitle: undefined,
+  },
+}
+
+export const TextContent: Story = {
+  args: {
+    extra: undefined,
+    subtitle: undefined,
+    title: undefined,
+  },
+}
+
+export const ActionContent: Story = {
+  args: {
+    children: (
+      <div style={{ display: 'flex', alignItems: 'center', justifyContent: 'space-between' }}>
+        {'Card Content'}
+        {actionButton}
+      </div>
+    ),
+    extra: undefined,
+    subtitle: undefined,
+    title: undefined,
+  },
+}
+
+export const TableContent: Story = {
+  args: {
+    children: <Table columns={columns} data={data} rowKey={rowKey} />,
+    extra: undefined,
+    subtitle: undefined,
+    title: undefined,
+  },
+}
+
+export const Loading: Story = {
+  args: {
+    isLoading: true,
+  },
+}
--- a/src/components/Card/Card.test.tsx
+++ b/src/components/Card/Card.test.tsx
@@ -1,0 +1,54 @@
+/**
+ * Copyright 2023 Mia srl
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { render, screen, waitFor } from '../../test-utils'
+import { Button } from '../Button'
+import { Card } from '.'
+
+describe('Card Component', () => {
+  const props = {
+    docLink: 'https://mia-platform.eu',
+    extra: <Button>{'Button'}</Button>,
+    subtitle: 'Card Subtitle',
+    title: 'Card Title',
+  }
+
+  beforeEach(() => {
+    jest.resetAllMocks()
+  })
+
+  test('renders card correctly', () => {
+    const { asFragment } = render(<Card {...props}>{'Card Content'}</Card>)
+    expect(asFragment()).toMatchSnapshot()
+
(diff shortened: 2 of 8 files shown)


src/components/Card/Card.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const docLinkIcon = useMemo(() => (
+    <Icon color={palette?.action?.link?.active} name="PiBookOpen" size={16} />
+  ), [palette?.action?.link?.active])
+
+  return (
+    <div className={card}>
+      <Skeleton
+        active
+        loading={isLoading}
+        paragraph={{ rows: 6, width: ['30%', '80%', '65%', '30%', '70%', '60%'] }}
```
Danielecina, line 59: Create a static reference outside of the component

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,86 @@
...
+        {(title || subtitle) && <div className={header}>
+          <div className={titles}>
+            {title && <div className={titleStyle}>
+              <H3>{title}</H3>
+              {docLink && <div className={icon}>
+                <Button
+                  icon={docLinkIcon}
+                  shape={Circle}
+                  type={Ghost}
+                  onClick={() => window.open(docLink, '_blank')}
```
Danielecina, line 70: I propose to use a useCallback here to avoid function reconstruction

src/components/Card/Card.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    expect(screen.getAllByRole('paragraph')).toHaveLength(1)
+    expect(screen.getByText(/Card Subtitle/i)).toBeVisible()
+
+    expect(screen.getByText(/Card Content/i)).toBeVisible()
+  })
+
+  test('renders skeleton correctly', async() => {
+    const { asFragment } = render(<Card isLoading={true} />)
+    await waitFor(() => expect(asFragment()).toMatchSnapshot())
```
Giogia, line 52:
```suggestion
    expect(asFragment()).toMatchSnapshot()
```

src/components/Card/Card.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { render, screen, waitFor } from '../../test-utils'
```
Giogia, line 19:
```suggestion
import { render, screen } from '../../test-utils'
```

Giogia (changes requested):
This waitfor should be necessary only for Ant components

src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+const actionIcon = <Icon color="#b91200" name="PiTrash" size={16} />
+const actionButton = (
+  <Button
+    hierarchy={Danger}
+    icon={actionIcon}
+    shape={Circle}
+    type={Outlined}
+  />
+)
```
Giogia, line 40: let's move these to a new mocks file.

src/components/Card/Card.tsx
Giogia: I would also have a chat with designers' team for the skeleton shape. This is a bit to big in my opinion, I would stick to 3-4 rows.

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,86 @@
...
+  return (
+    <div className={card}>
+      <Skeleton
+        active
+        loading={isLoading}
+        paragraph={{ rows: 6, width: ['30%', '80%', '65%', '30%', '70%', '60%'] }}
+      >
+        {(title || subtitle) && <div className={header}>
+          <div className={titles}>
+            {title && <div className={titleStyle}>
```
Giogia, line 63:
```suggestion
          <div className={heading}>
            {title && <div className={title}>
```

Giogia: I checked and header can be used as the container for the heading (that indicates only the typography instead)

RiccardoCorona-MiaPlatform: I cannot use `title` because it is already the name used by the prop 😬

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,89 @@
...
+      <Skeleton
+        active
+        loading={isLoading}
+        paragraph={skeletonParagraph}
+      >
+        {(title || subtitle) && <div className={header}>
+          <div className={titles}>
+            {title && <div className={titleStyle}>
+              <H3>{title}</H3>
+              {docLink && <div className={icon}>
```
Giogia, line 68:
```suggestion
              {docLink && <div className={styles.docLink}>
```

src/components/Card/Card.tsx
Giogia: Also, can the classname be applied directly to H3 element instead of wrapping it with a div?

src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+const meta = {
+  component: Card,
+  args: {
+    ...Card.defaultProps,
+    children: 'Card Content',
+    docLink: 'https://mia-platform.eu',
+    extra: <Button>{'Button'}</Button>,
+    subtitle: 'Card Subtitle',
+    title: 'Card Title',
```
Giogia, line 35: It would be nice to reuse the same props also in tests. I would create a mocks file with common definitions. 
```
export const extra = <Button>{'Button'}</Button>
```

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { Skeleton } from 'antd'
+
+import { Shape, Type } from '../Button/Button.types'
+import { BodyS } from '../Typography/BodyX/BodyS'
+import { Button } from '../Button'
+import { CardProps } from './Card.props'
+import { H3 } from '../Typography/HX/H3'
+import { Icon } from '../Icon'
+import { skeletonParagraph } from './Card.mocks'
+import styles from './Card.module.css'
```
Danielecina, line 29: Rename this styles in classes

src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+        {actionButton}
+      </div>
+    ),
+    extra: undefined,
+    subtitle: undefined,
+    title: undefined,
+  },
+}
+
+export const TableContent: Story = {
```
Giogia, line 101: Can we add some data such that the table scrolls?

src/components/Card/Card.tsx
RiccardoCorona-MiaPlatform: > Also, can the classname be applied directly to H3 element instead of wrapping it with a div?

Nope, because `className` is not a valid prop for our H3 component

src/components/Card/Card.stories.tsx
Giogia: I would also assign these to the Default story, so we can define other stories in the opposite way (Title story will have title prop instead of undefined props)

src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+import type { Meta, StoryObj } from '@storybook/react'
+
+import { columns, data, rowKey } from '../Table/Table.mocks'
+import { Button } from '../Button'
+import { Card } from '.'
+import { Table } from '../Table'
+import { actionButton } from './Card.mocks'
+
+const meta = {
+  component: Card,
```
Giogia, line 28: we should disable the extra control to avoid this

Giogia: image

src/components/Card/Card.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+    expect(screen.getByText(/Card Subtitle/i)).toBeVisible()
+
+    expect(screen.getByText(/Card Content/i)).toBeVisible()
+  })
+
+  test('renders skeleton correctly', () => {
+    const { asFragment } = render(<Card isLoading={true} />)
+    expect(asFragment()).toMatchSnapshot()
+  })
+})
```
Giogia, line 54: Can we add two tests only for heading and only for content? Snapshots will be enough

src/components/Card/Card.stories.tsx
RiccardoCorona-MiaPlatform: Should we do the same for `children` as well? Otherwise, the same problem occurs in content-related stories

RiccardoCorona-MiaPlatform: Disabling them, however, makes them disappear from the Storybook documentation, so I don't know if that is exactly what we want 😬

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const onClickDocLink = useCallback(() => window.open(docLink, '_blank'), [docLink])
+
+  return (
+    <div className={card}>
+      <Skeleton
+        active
+        loading={isLoading}
+        paragraph={skeletonParagraph}
+      >
```
Giogia, line 63: As suggested by the design team
```suggestion
        paragraph={{ rows: 3, width: ['80%', '65%', '70%'] }}
```

src/components/Card/Card.stories.tsx
RiccardoCorona-MiaPlatform: I found out the following:
- `argTypes: { foo: { table: { disable: true } } }` hides the prop `foo` from the whole UI ❌
- `argTypes: { foo: { control: false } }` hides the controls but does not remove the prop `foo` from the documentation ✅

Giogia (dismissed):

Giogia (approved):

src/components/Card/Card.mocks.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import { Hierarchy, Shape, Type } from '../Button/Button.types'
+import { Button } from '../Button'
+import { Icon } from '../Icon'
+
+const { Danger } = Hierarchy
+const { Circle } = Shape
+const { Outlined } = Type
+
+const actionIcon = <Icon color="#b91200" name="PiTrash" size={16} />
```
Giogia, line 29:
```suggestion
const actionIcon = (
  <Icon 
      color="#b91200" 
      name="PiTrash" 
      size={16} 
   />
)
```

Danielecina (approved):

Giogia (approved):

PR #310, opened 2024-02-13
fix(IDP-134): fix table border and default pagination

### LINKED ISSUE #291: Allow Table pagination to keep default values if not passed explicitly

In order to keep the default pagination the table now requires to pass 
```pagination={...Table.pagination, ...customPartialPagination}```

while it would be better to just pass
```pagination={customPartialPagination}```

## The description of the bug or the rationale of your proposal

Easier developer experience in using the Table component

### PR DESCRIPTION
**Pull request by Giogia**
### Description

Fix Table style and default pagination

fixes #291

##### Table

Added lateral border
Keep pagination default values if not overwritten

[IDP-134](https://makeitapp.atlassian.net/browse/IDP-134)

### Checklist

- [x] added the link to the Jira task
- [x] tests are included

### DIFF
--- a/src/components/Table/Table.test.tsx
+++ b/src/components/Table/Table.test.tsx
@@ -42,6 +42,11 @@ describe('Table Component', () => {
 
     expect(screen.getByRole('table')).toBeVisible()
 
+    expect(screen.getByText('Field 1')).toBeInTheDocument()
+    expect(screen.getByText('Field 2')).toBeInTheDocument()
+    expect(screen.getByText('Field 3')).toBeInTheDocument()
+    expect(screen.getByText('Field 4')).toBeInTheDocument()
+
     expect(screen.getByRole('row', { name: 'Value 1 Value 1 Value 1 Value 1' })).toBeVisible()
     expect(screen.getByRole('row', { name: 'Value 2 Value 2 Value 2 Value 2' })).toBeVisible()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/components/Table/Table.tsx [hunk 1]
```diff
@@ -65,7 +65,7 @@ export const Table = <RecordType extends GenericRecord>({
         footer={footer}
         loading={false}
         locale={intlLocale}
-        pagination={pagination}
+        pagination={{ ...Table.pagination, ...pagination }}
```
Danielecina, line 68: Serve davvero? Nelle default props tanto le passiamo, se l'utente decide che non le vuole glielo lascerei fare?

RiccardoCorona-MiaPlatform (dismissed):

src/components/Table/Table.tsx
fredmaggiowski, line 68: Could a `useMemo` be helpful?

Giogia: This is the fix requested in #291 . 

In this way, we allow to provide only a partial default pagination, overwriting only the necessary fields while maintaining the other default values. From the user's perspective, there is no need to spread Table.pagination explicitly.

Giogia: Added ✅

Danielecina: ok, I see that the code is changed 👍

Danielecina (approved):

RiccardoCorona-MiaPlatform (approved):

PR #332, opened 2024-02-20
fix: Typography margin & Card header

RiccardoCorona-MiaPlatform:
##### Typography

Removed margin from both BodyX and HX components.

##### Card

Card header adjusted to support long titles by going into ellipsis.


src/components/Table/Table.module.css
```diff
@@ -1,15 +1,33 @@
+/**
```
Danielecina, line 1: Why didn't the pipeline license check flag this file? 
do we check that there is the command on the pipeline that calls the script to check?

src/components/Typography/BodyX/BodyX.stories.tsx
```diff
@@ -73,14 +73,14 @@ export const All = {
     children: { table: { disable: true } },
   },
   render: (args: object) => (
-    <>
+    <div style={{ display: 'flex', flexDirection: 'column', gap: 16 }}>
```
Danielecina, line 76: Can we use outside a shared style?

Giogia, line 76: I see this is a repeating pattern. 
Can we create some mocks (e.g. one long text and one standard length) to be reused in tests and stories?

src/components/Typography/BodyX/BodyX.module.css
```diff
@@ -16,6 +16,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+.body {
```
Giogia, line 19: Can we apply this style using Ant design tokens?

src/components/Typography/HX/HX.module.css
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.h {
+  margin: 0 !important;
```
Giogia, line 20: Can we apply this style using Ant design tokens?

src/components/Card/Card.module.css
```diff
@@ -38,12 +39,14 @@
...
   justify-content: space-between;
   padding: calc(var(--spacing-padding-sm, 8) * 1px) calc(var(--spacing-padding-xl, 24) * 1px);
+  gap: calc(var(--spacing-gap-xl, 24) * 1px);
 }
 
 .heading {
   display: flex;
   flex-direction: column;
   gap: calc(var(--spacing-gap-sm, 8) * 1px);
+  overflow-x: hidden;
```
Danielecina, line 49: why we need to use these the overflow-x?

src/components/Typography/BodyX/BodyX.stories.tsx
Giogia, line 76: I would either extract a common style 
```
{ display: 'flex', flexDirection: 'column', gap: 16 }
```
or a common decorator
```
({children}) => (
    <div style={{ display: 'flex', flexDirection: 'column', gap: 16 }}>
    {children}
    </div>
    ```
    to be put in mocks file and reused in different stories. Same applies also for HX

File: src/components/Card/Card.tsx [hunk 6]
```diff
@@ -63,7 +63,7 @@ export const Card = ({
         {(title || subtitle) && <div className={header}>
           <div className={heading}>
             {title && <div className={styles.title}>
-              <H3>{title}</H3>
+              <H3 ellipsis={{ rows: 1, tooltip: title }}>{title}</H3>
```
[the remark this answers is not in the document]
RiccardoCorona-MiaPlatform: We are already just passing `ellipsis`, however, `ellipsis` needs a specific number of rows and something to show in the tooltip. Without specifying `rows`, the component doesn't know when it should truncate, and without specifying `tooltip`, it doesn't know what to show you on mouse hover (and therefore doesn't show you anything)

src/components/Table/Table.module.css
RiccardoCorona-MiaPlatform: I do not know where the pipeline instructions are, however, I noticed that the absence of the comment with the license does not result in any error 😬

src/components/Typography/HX/HX.module.css
RiccardoCorona-MiaPlatform: Done ✅

src/components/Typography/BodyX/BodyX.module.css
RiccardoCorona-MiaPlatform: For paragraphs, there are no ad hoc display tokens. However, having used those for titles, here I put `margin: 0 !important` inside the individual `bodyS`, `bodyM`, `bodyL` so as not to create more classes than necessary

src/components/Typography/BodyX/BodyX.stories.tsx
RiccardoCorona-MiaPlatform: I extracted the `displayAll` style ✅

src/components/Card/Card.module.css
RiccardoCorona-MiaPlatform: We need it to prevent the title from breaking through the header and the card itself:

Screenshot from 2024-02-20 13-31-57

src/components/Typography/BodyX/BodyX.module.css
```diff
@@ -17,16 +17,19 @@
  */
 
 .bodyS {
+  margin: 0 !important;
```
Giogia, line 20: Are these still required?

Giogia (approved):

src/components/Typography/BodyX/BodyX.module.css
RiccardoCorona-MiaPlatform: Yes, because `titleMarginBottom` and `titleMarginTop` do not affect paragraphs (and thus our bodies)

Giogia: yes, they  are

src/components/Table/Table.module.css
Giogia: We can only run the script and see if it adds any license. Unfortunately at the moment, we are not able to check if some files need to be modified from the pipeline

PR #334, opened 2024-02-20
feat(IDP-68): Modal component & useModal hook

RiccardoCorona-MiaPlatform:
##### Modal

Modal component implementation.

##### useModal

useModal hook implementation.

[Creazione componente Modale](https://makeitapp.atlassian.net/browse/IDP-68)

### Checklist

- [X] added the link to the Jira task
- [x] tests are included
- [X] you are not committing extraneous files or sensible data


src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ *
+ *
+ * @link https://ant.design/components/modal
+ * @returns {Modal} Modal component
+ */
+export const Modal = ({
+  children,
+  docLink,
```
Danielecina, line 58: è un po' custom sul nostro uso in console come prop. mi torna la sua presenza però astrarrei il concetto.
Userei qualcosa tipo iconName che è una implementazione delle nostre icone e un iconLink?

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,216 @@
...
+        </>
+      )
+    }
+
+    return (
+      <div className={content}>
+        {children}
+      </div>
+    )
+  }, [children, extension, isExtensionOpen, modalContentWrapperClassNames, size])
```
Danielecina, line 190: il children è una dipendenza che cambia sempre no?

src/components/Modal/Modal.types.ts
```diff
@@ -0,0 +1,46 @@
...
+  labelClose?: string,
+
+  labelOpen?: string,
+
+  title?: ReactNode,
+}
+
+export type Footer = {
+
+  // TODO: documentare il fatto che ci andrebbero dei button
```
Danielecina, line 42: .

src/hooks/useModal/useModal.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { useCallback, useState } from 'react'
```
Danielecina, line 19: perchè fare un file a parte? dentro questa cartella mi aspetto che ci sia solo lui

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,216 @@
...
+/**
+ *
+ *
+ * @link https://ant.design/components/modal
+ * @returns {Modal} Modal component
+ */
+export const Modal = ({
+  children,
+  docLink,
+  extension,
```
Danielecina, line 59: lo farei più espressivo. è un sider un aside della modale.

src/hooks/useModal/useModal.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { ModalAPI } from './useModal.types'
+
+export const useModal = (): ModalAPI => {
+  const [isModalVisible, setIsModalVisible] = useState(false)
+
+  const openModal = useCallback(() => setIsModalVisible(true), [setIsModalVisible])
+  const closeModal = useCallback(() => setIsModalVisible(false), [setIsModalVisible])
```
Danielecina, line 27: sn fn usiamo la sintassi onXXX.

src/hooks/useModal/useModal.ts
```diff
@@ -0,0 +1,31 @@
...
+  const changeModalStatus = useCallback(() => setIsModalVisible(!isModalVisible), [setIsModalVisible, isModalVisible])
```
Danielecina, line 28: Questo è un uso scorretto del setState. 
Il set state è asincrono, se bindi così il vecchio visible potrebbe non avere il valore che ti aspetti.
Per avere la certezza di sapere quale l'attuale valore nel tuo state devi usare questa sintassi

```suggestion
  const changeModalStatus = useCallback(() => setIsModalVisible(prevState => !prevState), [])
```

src/components/Modal/Modal.tsx
RiccardoCorona-MiaPlatform: Not necessarily, we have modals whose content is only informative. In addition, components such as Ant's Input or Form do not generate a change of `children` upon, for example, text entry or the flag of a checkbox

src/components/Modal/Modal.types.ts
RiccardoCorona-MiaPlatform: Done ✅

src/hooks/useModal/useModal.ts
RiccardoCorona-MiaPlatform: We also have the tests and mocks files within the folder

src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+export const WithNoHeader: Story = {
+  args: {
+    ...meta.args,
+    title: undefined,
+  },
+  decorators: [(_, { args }) => <WithOpenButton {...args} />],
+}
+
+export const WithNoFooter: Story = {
```
fredmaggiowski, line 79: This story is to provide an example of a modal that has no custom footer but displays the empty footer instead, right? Can we express this better in the story name? otherwise I'd expect to see a modal with no footer at all

src/hooks/useModal/useModal.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+
+import { Button, Modal } from '../..'
+import { useModal } from '.'
+
+describe('useModal', () => {
```
Giogia, line 24: I would also add unit tests for this hook:

```suggestion
describe('useModal', () => {
  beforeEach(() => {
    jest.clearAllMocks()
  })

  it('returns isModalVisible, openModal, closeModal and toggleModal properties', async () => {
    const {result} = renderHook(() => useModal())

    const {isModalVisible, openModal, closeModal, toggleModal} = result.current
    expect(isModalVisible).toEqual(false)

    openModal()
    expect(result.current.isModalVisible).toEqual(true)

    closeModal()
    expect(result.current.isModalVisible).toEqual(false)

    toggleModal()
    expect(result.current.isModalVisible).toEqual(true)

    toggleModal()
    waitFor(() => expect(result.current.isModalVisible).toEqual(false))
  })
```

src/hooks/useModal/index.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export { useModal } from './useModal'
```
Giogia, line 19: we need to export this hook from hooks/index.ts

src/components/Modal/Modal.tsx
fredmaggiowski: i agree to keep the docLink concept within the modal; we might add a more customizable link in the future but as of now it hardly makes sense to keep it outside the component

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,213 @@
...
+    isBodyFullWidth && bodyFullWidth,
+  ]), [isBodyFullWidth])
+  const modalContentWrapperClassNames = useMemo(() => classNames([
+    content,
+    size !== Small && aside && contentWrapper,
+  ]), [aside, size])
+
+  const [isAsideOpen, setIsAsideOpen] = useState(false)
+
+  const modalTitle = useMemo(() => (
```
fredmaggiowski, line 94: these memoized components could be full fledged components, this would make evolving them easier and this Modal codebase far more shorter

src/hooks/useModal/useModal.ts
```diff
@@ -0,0 +1,37 @@
...
+ * for opening, closing, and changing the state of the modal.
+ */
+export const useModal = (): ModalAPI => {
+  const [isModalVisible, setIsModalVisible] = useState(false)
+
+  const openModal = useCallback(() => setIsModalVisible(true), [])
+  const closeModal = useCallback(() => setIsModalVisible(false), [])
+  const changeModalStatus = useCallback(() => setIsModalVisible(prevState => !prevState), [])
+
+  return { isModalVisible, openModal, closeModal, changeModalStatus }
```
fredmaggiowski, line 36: rather than "changeModalStatus", I'd call the function `toggleModal` which is far more standard

```suggestion
  return { isModalVisible, openModal, closeModal, toggleModal }
```

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,213 @@
...
+  isClosable,
+  isVisible,
+  onCloseClick,
+  size,
+  title,
+}: ModalProps): ReactElement => {
+  const { palette } = useTheme()
+
+  const docLinkIcon = useMemo(() => (
+    <Icon color={palette?.action?.link?.active} name="PiBookOpen" size={16} />
```
Giogia, line 72:
```suggestion
    <Icon color={palette?.action?.link?.active} name="PiBookOpen" size={shape?.size?.md} />
```

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,213 @@
...
+  docLink,
+  footer,
+  isBodyFullWidth,
+  isClosable,
+  isVisible,
+  onCloseClick,
+  size,
+  title,
+}: ModalProps): ReactElement => {
+  const { palette } = useTheme()
```
Giogia, line 69:
```suggestion
  const { palette, shape } = useTheme()
```

src/hooks/useModal/useModal.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { ModalAPI } from './useModal.types'
+
+/**
+ * A hook that allows the use of Modal components.
+ *
+ * @returns {ModalAPI} An object which includes the state of the modal (visible or hidden) and methods
+ * for opening, closing, and changing the state of the modal.
+ */
+export const useModal = (): ModalAPI => {
```
fredmaggiowski, line 29: The current implementation of the useModal we have in our internal repositories is actually a little bugous, since it requires to have both the modal and the hook instantiated and requires the developer to correctly map the modal visible state with the desired modal.

For example I'm always required to something like


```jsx
import { Modal, useModal } from ...

function MyComp({}) {
   const { isModalVisible, open, close, toggle } = useModal()

   return <Modal visible={isModalVisible} />
}
```

It would be nice to have some kind of interface that helps managing such visibility state without the need for the developer remapping.

What if the hook returns the Modal component as well?

So I'd do something like:

```jsx
import { useModal } from ...

function MyComp({}) {
   const { ModalComponent, open, close, toggle } = useModal()

   return <ModalComponent />
}
```

fredmaggiowski (changes requested):

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,213 @@
...
+            {buttons}
+          </div>
+          {extra}
+        </>}
+      </div>
+    )
+  }, [footer])
+
+  const modalContent = useMemo(() => {
+    if (size !== Small && aside) {
```
Giogia, line 125: I think it is easier to understand if we invert the condition:

```suggestion
    if (size == Small || !aside) {
      return (
      <div className={content}>
        {children}
      </div>
    )}
```

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,213 @@
...
+            shape={Circle}
+            type={Ghost}
+            onClick={onClickDocLink}
+          />
+        </div>}
+      </>}
+    </div>
+  ), [docLink, docLinkIcon, onClickDocLink, title])
+
+  const modalFooter = useMemo(() => {
```
Giogia, line 110: Can we move footer definition after content definition ? 😁

src/components/Modal/Modal.mocks.tsx
```diff
@@ -0,0 +1,166 @@
...
+)
+
+export const childrenLong = (
+  <>
+    <Table
+      columns={columns}
+      data={data}
+      rowKey={rowKey}
+    />
+    {`Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque vel dolor ac sem congue ornare rhoncus quis tortor.
```
Giogia, line 65: Should we move this string to a more generic utils file? So every component can use it

RiccardoCorona-MiaPlatform: Let's rather introduce an ad-hoc library to generate lorem ipsum texts (I'd address this in a dedicated PR) 🤙🏼

src/components/Modal/Modal.stories.tsx
RiccardoCorona-MiaPlatform: I called it `WithEmptyFooter` (and renamed the other one to `WithEmptyHeader`) ✅

src/components/Modal/Modal.tsx
RiccardoCorona-MiaPlatform: The prop `size` accepts only number-specific values, and neither the use of `parse` nor `as const` convert the theme variable to an acceptable type 😬

src/components/Modal/Modal.tsx
RiccardoCorona-MiaPlatform: Not needed anymore ✅

src/components/Modal/Modal.tsx
RiccardoCorona-MiaPlatform: I converted this condition to a const called `hasAside` which I put at the top of the file to facilitate the assignment of `classNames`. In this new scenario it seems more useful to me to keep it as is (otherwise I would have to call the const `doesNotHaveAside`)

src/components/Modal/Modal.module.css
```diff
@@ -0,0 +1,225 @@
...
+      gap: calc(var(--spacing-gap-sm, 8) * 1px);
+      overflow-y: auto;
+    }
+
+    .asideLabel {
+      display: flex;
+      align-items: center;
+      cursor: pointer;
+      gap: calc(var(--spacing-padding-md, 8) * 1px);
+      color: var(--palette-action-link-active, #005dc9);
```
Giogia, line 122:
```suggestion
      color: var(--palette-action-link-active, #005dc9);
      transition: color 0.2s ease-in-out;
```

src/components/Modal/Modal.module.css
```diff
@@ -0,0 +1,225 @@
...
+  :global(.mia-platform-modal-body) {
+    height: calc(100% - 2 * calc(var(--shape-size-xl, 32) * 1px));
+    padding: calc(var(--spacing-padding-lg, 16) * 1px) 0;
+
+    .body {
+      height: 100%;
+      display: grid;
+      grid-template-columns: 1fr;
+      padding: 0 calc(var(--spacing-padding-xl, 24) * 1px);
+      transition: grid-template-columns 250ms ease-in-out;
```
Giogia, line 69:
```suggestion
      transition: grid-template-columns 0.2s ease-in-out;
```

src/components/Modal/Modal.module.css
```diff
@@ -0,0 +1,225 @@
...
+    .bodyFullWidth {
+      padding: 0 !important;
+    }
+
+    .bodyWithAsideClosed {
+      grid-template-columns: 100% 35%;
+    }
+    .bodyWithAsideClosed .aside {
+      display: none;
+      animation: 250ms fadeOut;
```
Giogia, line 82:
```suggestion
      animation: 0.2s fadeOut;
```

src/components/Modal/Modal.module.css
```diff
@@ -0,0 +1,225 @@
...
+      padding: calc(var(--spacing-padding-lg, 16) * 1px);
+      gap: calc(var(--spacing-gap-sm, 8) * 1px);
+      overflow-y: auto;
+    }
+
+    .asideLabel {
+      display: flex;
+      align-items: center;
+      cursor: pointer;
+      gap: calc(var(--spacing-padding-md, 8) * 1px);
```
Giogia, line 121:
```suggestion
      gap: calc(var(--spacing-padding-xs, 4) * 1px);
```

src/components/Modal/Modal.module.css
```diff
@@ -0,0 +1,225 @@
...
+      width: 100%;
+      height: 100%;
+      display: flex;
+      flex-direction: column;
+      gap: calc(var(--spacing-gap-xl, 24) * 1px);
+      overflow-y: auto;
+    }
+
+    .contentWrapper {
+      gap: calc(var(--spacing-gap-md, 8) * 1px) !important;
```
Giogia, line 103:
```suggestion
      gap: calc(var(--spacing-gap-sm, 8) * 1px) !important;
```

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,96 @@
...
+  ]), [size])
+
+  return (
+    <AntModal
+      centered
+      className={modalClassNames}
+      closable={isClosable}
+      footer={<Modal.Footer footer={footer} />}
+      keyboard={isClosable}
+      maskClosable={false}
```
fredmaggiowski, line 71: I'd map this prop to the isClosable flag.

Danielecina: closable is used to show the X to close modal. 
the prop quoted to you is already mapped below as` closable={isClosable}`

fredmaggiowski: yes i know, I'd map `maskClosable` as well, a modal that is not closable by clicking outside of it is extremely annoying.

Danielecina: I check this with the designer and @Giogia anche this props needs to be false

fredmaggiowski: This design choice makes very little sense to me. Let's keep it this way for now but i'll open an issue for this.

PR #358, opened 2024-02-29
chore(pkg): add support for `exports` field

amountainram:
### Description

To allow tree-shaking in latest vite.js versions (and other tools), it is mandatory to seek `ESM` libraries using the [`exports` field](https://nodejs.org/api/packages.html#subpath-exports).

If not specified, libraries will default back to the `main` field which is usually the `cjs` bundle, thus preventing tree-shaking.

According to the `module` field, this commit adds a meaningful support for the `exports` field

##### [INVOLVED_COMPONENT_HERE]

none

### Checklist

- [x] tests are included (none needed)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,12 @@
   "main": "dist/cjs/console-design-system-react.cjs",
   "module": "dist/es/console-design-system-react.js",
   "types": "dist/index.d.ts",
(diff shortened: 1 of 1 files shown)


fredmaggiowski (approved):
LGTM

PR #397, opened 2024-03-18
fix: define fixed height of feedbackMessage component to have aligned icons

ThisIsDemetrio:
##### [FeedbackMessage]

Component will have an fixed height of 40px instead of relying content (which caused the height to change based on the icon shown).

Moreover, several CSS instructions have been included to override ant style in order to have the icon aligned with the text inside.

### Checklist

- [ ] tests are included
- [X] you are not committing extraneous files or sensible data


fredmaggiowski (approved):
fragile 🥶

src/hooks/useFeedbackMessage/useFeedbackMessage.module.css
```diff
@@ -50,10 +50,18 @@
...
+    align-items: center;
     padding: calc(var(--spacing-padding-md, 8) * 1px) !important;
+    display: inline-flex !important;
+    height: 40px;
 
     [class~=ant-message-custom-content] {
       display: flex !important;
+      align-items: center;
+
+      [class~=anticon-loading] {
```
Giogia, line 62: We should try to achieve the same behavior using Ant Design Tokens instead.
https://ant.design/components/message#design-token

Giogia: we tried but it's easier this way

Giogia (approved):

PR #464, opened 2024-05-23
feat: add Breadcrumb component

fedepini:
### Description

This PR has the goal of adding the `Breadcrumb` component to ease navigation between sections.

Screenshot 2024-06-12 at 16 30 50

### Checklist

- [x] tests are included


epessina (approved):

fredmaggiowski (changes requested):
I dind't finish the review yet, I'm still missing a few files

src/components/Breadcrumb/Breadcrumb.module.css
```diff
@@ -0,0 +1,253 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.breadcrumb {
+  /* TODO: these CSS variables are missing in DS */
```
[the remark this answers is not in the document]
FrankieBortot: On Figma we have the semantic layer (--palette-background-neutral), but we miss the "component layer", where we should define each component token. It's a task we have in backlog, but it has low priority since it has low value. We can discuss this decision and start adding this layer!

src/components/Breadcrumb/assets/caret-full-down.svg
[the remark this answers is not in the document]
FrankieBortot: Caret-full has been usedto be distinguished from from the outline-caret used to separate esch level of the breadcrumb; using the same caret could be misleading and confusing

src/components/Breadcrumb/Breadcrumb.utils.tsx
```diff
@@ -0,0 +1,97 @@
...
+      isHidden={ctx.isHidden}
+      isLastItem={idx === (itemList?.length ?? 1) - 1}
+      isLoading={ctx.isLoading}
+      item={typedItem}
+      key={typedItem?.key ?? `breadcrumb-item-${idx}`}
+    />
+  )
+}
+
+export const computeItems = (
```
fredmaggiowski, line 58: This function is responsibile for deciding how many items should be visible based on the container width, am I right?

Can we give it a more explicative name? 🤔 

I think of something like `getItemsFittingContainer` but I'm not convinced, yet.

src/components/Breadcrumb/Breadcrumb.utils.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const nextVisibleItems: BreadcrumbButton[] = []
+  const nextCollapsedItems = items.slice(1, items.length - 1)
+
+  const elements = Array.from(hiddenContainer.children)
+  let currWidth = 0
+
+  nextVisibleItems.push(items[items.length - 1])
+
+  currWidth += elements[elements.length - 1].getBoundingClientRect().width
```
fredmaggiowski, line 75: Are these accesses to elements safe? Can `elements` have length `0`? if so this line would throw a TypeError 🤔

src/components/Breadcrumb/Breadcrumb.utils.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const separatorWidth = BREADCRUMB_SEPARATOR_SIZE + (BREADCRUMB_SEPARATOR_PADDING * 2)
+
+  const nextVisibleItems: BreadcrumbButton[] = []
+  const nextCollapsedItems = items.slice(1, items.length - 1)
+
+  const elements = Array.from(hiddenContainer.children)
+  let currWidth = 0
+
+  nextVisibleItems.push(items[items.length - 1])
```
fredmaggiowski, line 73: Is this access to items safe? can items have length 0?

src/components/Breadcrumb/Breadcrumb.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { BreadcrumbItemMenu } from './Breadcrumb.types'
+
+const meta = {
+  component: Breadcrumb,
+} satisfies Meta<typeof Breadcrumb>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
```
fredmaggiowski, line 35: When running the storybook, it will display a code sample for each story, since I've seen some stories feature an unintelligible codeblock. How are these stories code samples displayed? are they usable from a storybook reader?

src/components/Breadcrumb/Breadcrumb.Dropdown.tsx
```diff
@@ -0,0 +1,113 @@
...
+  return (
+    <div className={styles.dropdownMenuContainer}>
+      {
+        Boolean(item.menu?.search) && (
+          <div className={styles.dropdownMenuSearch}>
+            <Input
+              allowClear={getSearchOption(item.menu?.search, 'allowClear')}
+              autoFocus
+              placeholder={getSearchOption(item.menu?.search, 'placeholder') ?? 'Search...'}
+              // @ts-expect-error size 12 is not accepted by Icon component but supported by underling SVG
```
fredmaggiowski, line 71: Why a 12px icon? @FrankieBortot

src/components/Breadcrumb/Breadcrumb.Dropdown.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  return (
+    <div className={styles.dropdownMenuContainer}>
+      {
+        Boolean(item.menu?.search) && (
+          <div className={styles.dropdownMenuSearch}>
+            <Input
+              allowClear={getSearchOption(item.menu?.search, 'allowClear')}
+              autoFocus
+              placeholder={getSearchOption(item.menu?.search, 'placeholder') ?? 'Search...'}
```
fredmaggiowski, line 70: `Search...` should be internazionalized

src/components/Breadcrumb/Breadcrumb.Dropdown.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    return item.menu.items.filter((subItem) => subItem.label?.toLowerCase().includes(searchValue.toLowerCase()))
+  }, [item, searchValue])
+
+  const menuItems = filteredItems.map<ItemType>((menuItemData, currentIndex) => {
+    return {
+      key: buildMenuItemKey(menuItemData, currentIndex),
+      icon: menuItemData?.icon,
+      label: (
+        <BodyS ellipsis={{ rows: 1, tooltip: menuItemData?.label }}>
```
fredmaggiowski, line 55: this prop should be memoized, possibily the whole `filteredItems.map` result

src/components/Breadcrumb/Breadcrumb.module.css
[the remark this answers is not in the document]
fredmaggiowski: I'd open an issue to start this discussion and then we fix this TODO then, ok?

@fedepini can you open it yourself?

src/components/Breadcrumb/Breadcrumb.Dropdown.tsx
FrankieBortot: Is it the close-button?
In that case, Antd already uses a close-button (an X inside a filled circle) with dimension 12px.

fredmaggiowski: It seems to be the search icon: `PiMagnifyingGlass`

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,99 @@
...
+ * @returns {Breadcrumb} Breadcrumb component
+ */
+export const Breadcrumb = ({ isLoading, items, getPopupContainer }: BreadcrumbProps): ReactElement => {
+  const [visibleItems, setVisibleItems] = useState<BreadcrumbButton[]>([])
+
+  const breadcrumbRef = useRef<HTMLDivElement>(null)
+  const hiddenContainerRef = useRef<HTMLDivElement>(null)
+
+  const getDropdownContainer = useCallback(() => {
+    /* istanbul ignore next */
```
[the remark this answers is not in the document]
epessina: It is an edge case that should never happen and is not reproducible
The container is always mounted, therefore the ref should always be defined at this point. However, for type safety (and to avoid weird underlying behaviour breaking the UI), I prefer to check rather than cast.

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,99 @@
...
+  // Computes visible and collapsed items
+  useEffect(() => {
+    if (!items || items.length <= 2) {
+      setVisibleItems(items ?? [])
+      return
+    }
+
+    const container = breadcrumbRef.current
+    const hiddenContainer = hiddenContainerRef.current
+    /* istanbul ignore if  */
```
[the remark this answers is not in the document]
epessina: As previous conversation

src/components/Breadcrumb/Breadcrumb.props.ts
```diff
@@ -0,0 +1,67 @@
...
+   *       `React.ChangeEventHandler<HTMLInputElement>`
+   *       - allowClear: if true, allows clearing of the search input. <br> `boolean`
+   *       - placeholder: placeholder text for the search input. <br> `string`
+   *     - items: list of menu items in the breadcrumb. <br>
+   *     `object[]`:
+   *       - icon: icon to be displayed alongside the menu item. <br> `ReactNode`
+   *       - key: unique key for the menu item. Defaults to item index. <br> `string`
+   *       - label: label of the menu item. <br> `string`
+   */
+  items?: BreadcrumbItemType[];
```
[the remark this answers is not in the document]
epessina: It is debatable. From a functional point of view no, the component would be empty and useless. However, it is technically possible to pass no items to the component, so I took into consideration the possibility to properly handle it
We could make the property required, to hint the user, but that would not cover the case in which an empty array is passed, so I don't know if it is that useful

src/components/Breadcrumb/Breadcrumb.module.css
```diff
@@ -0,0 +1,253 @@
...
+            background-color: var(--palette-action-alternate-primary-active, #1890ff14) !important;
+          }
+        }
+
+        &.noInteraction {
+          cursor: default !important;
+          background-color: white !important;
+        }
+
+        & :global(.mia-platform-dropdown-menu-title-content) {
```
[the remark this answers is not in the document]
epessina: The global is stripped at compile time, so the space is needed because the selector targets child elements with the class, and not an element having both classes

src/components/Breadcrumb/Breadcrumb.utils.tsx
epessina: `elements` cannot have a length of less than 2, given the implementation of the component and the check on the `items` array at the beginning of breadcrumb `useEffect`.
However, for safety, I'll add a check on the length of the array also here

src/components/Breadcrumb/Breadcrumb.utils.tsx
epessina: As previous conversation

src/components/Breadcrumb/Breadcrumb.stories.tsx
epessina: Unfortunately storybook does not infer stories arguments if `decorators` are used (which is needed to properly open dropdowns). We have the same problem in the Modal component stories. We need to either review the pattern of having mocks in a separate file, or find a way to better tell storybook how the code blocks are made

src/components/Breadcrumb/Breadcrumb.Dropdown.tsx
epessina: I think it can be considered an Antd internal, since it's the `suffix` of an `Input` component. 12px seems to be the correct size for such an element

src/components/Breadcrumb/Breadcrumb.Dropdown.tsx
epessina: That's what `item.menu.search.placeholder` is there for. If we want to offer internal translation sheets we should set up them for the entire system

src/components/Breadcrumb/Breadcrumb.utils.tsx
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+import { ReactElement } from 'react'
+
+import { BREADCRUMB_COLLAPSED_WIDTH, BreadcrumbCollapsed } from './Breadcrumb.Collapsed'
+import { BREADCRUMB_SEPARATOR_PADDING, BREADCRUMB_SEPARATOR_SIZE } from './Breadcrumb.Separator'
+import { BreadcrumbButton, BreadcrumbItemMenuItem, BreadcrumbItemType } from './Breadcrumb.types'
+import { BreadcrumbItem } from './Breadcrumb.Item'
+
+export const renderItem = (
```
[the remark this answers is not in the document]
epessina: No real reasons, I simply prefer it like that. It should not have an impact either way

src/components/Breadcrumb/Breadcrumb.module.css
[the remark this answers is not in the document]
epessina: Opened it: https://github.com/mia-platform/design-system/issues/497

src/components/Breadcrumb/Breadcrumb.props.ts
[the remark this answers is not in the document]
fredmaggiowski: i agree with what you say; I'd still define it required to hint the user; there is still the chance for it to be empty but at least the component interface will make more sense :)

src/components/Breadcrumb/Breadcrumb.stories.tsx
fredmaggiowski: can you plesase open an issue on this matter?

src/components/Breadcrumb/Breadcrumb.Dropdown.tsx
fredmaggiowski: @Danielecina did you already consider i18n management?

src/components/Breadcrumb/Breadcrumb.utils.tsx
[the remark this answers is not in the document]
fredmaggiowski: I understand, making it a react component would allow us to use it with jsx syntax and to use hooks (e.g. for memoization) in case of need 🤔  

btw either way is fine, you can resolve this discussion if don't find it useful :)

src/components/Breadcrumb/Breadcrumb.stories.tsx
epessina: Done: https://github.com/mia-platform/design-system/issues/503

src/components/Breadcrumb/Breadcrumb.Dropdown.tsx
epessina: I have opened an issue on the matter https://github.com/mia-platform/design-system/issues/504

src/components/Breadcrumb/Breadcrumb.utils.tsx
epessina: Renamed it

src/components/Breadcrumb/assets/caret-full-down.svg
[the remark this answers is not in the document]
fredmaggiowski: > If we switch to the new Icon system (as per #477), we could include this version of the caret in our own package

Sounds great, I agree

src/components/Breadcrumb/Breadcrumb.Dropdown.tsx
fredmaggiowski: shall we include such size after we start managing icons as in #477 ?

fredmaggiowski (dismissed):
LGTM

epessina (dismissed):

LiliumCandidum (dismissed):

epessina (approved):

LiliumCandidum (approved):

PR #475, opened 2024-05-31
feat(useAntTheme): exposed generated ant theme

epessina:
### Description

This PR aims at exposing the generated [Ant Design theme](https://ant.design/docs/react/customize-theme#api) through a new `useAntTheme` hook.

Exposing the generated Ant theme is useful when there is the need to mix design system components with native Ant components.
The generated theme can be passed to an Ant [ConfigProvided](https://ant.design/components/config-provider) to keep a coherent styling between the two libraries.

### Checklist

- [x] tests are included

### DIFF
--- a/src/hooks/useAntTheme/index.ts
+++ b/src/hooks/useAntTheme/index.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2023 Mia srl
+ * Copyright 2024 Mia srl
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/hooks/useAntTheme/index.ts
+++ b/src/hooks/useAntTheme/index.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2024 Mia srl
+ * Copyright 2023 Mia srl
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,14 +16,4 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { useAntTheme } from './useAntTheme'
-import { useFeedbackMessage } from './useFeedbackMessage'
-import { useModal } from './useModal'
-import { useTheme } from './useTheme'
-
-export default {
-  useFeedbackMessage,
-  useModal,
-  useTheme,
-  useAntTheme,
-}
+export { useAntTheme } from './useAntTheme'
--- a/src/hooks/useAntTheme/useAntTheme.test.tsx
+++ b/src/hooks/useAntTheme/useAntTheme.test.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2023 Mia srl
+ * Copyright 2024 Mia srl
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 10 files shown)


fredmaggiowski:
Hi, can you further explain why this change is needed and what value brings to the lib end users?

src/hooks/useTheme/useTheme.ts
```diff
@@ -29,8 +31,11 @@ const { lightTheme: defaultTheme } = themes
...
  */
-export const useTheme = (): Theme => {
+export const useTheme = (): { theme: Theme, antTheme: ThemeConfig } => {
   const theme = useContext(ThemeContext)
 
-  return useMemo(() => theme ?? defaultTheme, [theme])
+  const themeWithFallback = useMemo(() => theme ?? defaultTheme, [theme])
+  const antTheme = useMemo(() => generateAntTheme(themeWithFallback), [themeWithFallback])
+
+  return { theme: themeWithFallback, antTheme }
```
fredmaggiowski, line 40: It feels odd for the `useTheme` to return an object holding a `theme` property. I'd expect the hook to return the actual theme (as in current implementation).


Also providing the underlying antd theme as our public interface doesn't feel like a sound choice. Can we provide an additional hook that returns the antdTheme so that only when the it is specifically needed you'd use the new hook instead? This would also make the change non-breaking (not that it is a problem since we are at v0, but this would still require some effort on all the existing codebase using it so I think it must be thought thouroughly)

epessina: Sure, I have re-implemented the feature as a new hook and changed the specifications of the PR

fedepini (dismissed):

src/hooks/useAntTheme/index.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright 2023 Mia srl
```
fredmaggiowski, line 2:
```suggestion
 * Copyright 2024 Mia srl
```

src/hooks/useAntTheme/useAntTheme.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* eslint-disable func-names */
```
fredmaggiowski, line 19: why these disabling? can you follow the lint rules (or update them to better suite our needs)?

src/hooks/useAntTheme/useAntTheme.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * Copyright 2023 Mia srl
```
fredmaggiowski, line 2:
```suggestion
 * Copyright 2024 Mia srl
```

fredmaggiowski (changes requested):

src/hooks/useAntTheme/useAntTheme.test.tsx
epessina: Done, and corrected `useTheme.test.tsx` too

fredmaggiowski: thanks, didn't notice it was implemented this way also for the other file 🤦

fredmaggiowski (approved):
LGTM

fedepini (approved):

PR #476, opened 2024-05-31
feat(themeProvider): add unit to css variables

epessina:
### Description

This PR aims to add the unit (`px`) to the CSS variables computed by the `ThemeProvider` component

The change is **breaking** for anyone who is using a syntax like:
```css
padding: calc(var(--spacing-padding-lg) * 1px);
```
since CSS `calc` [does not support multiplication between two values with unit](https://www.quora.com/Why-cant-you-use-CSS-calc-to-times-two-arguments-that-are-both-of-unit-type-pixel).

The code above should be transformed to:
```css
padding: var(--spacing-padding-lg);
```

##### Theme Provider

- Added unit to injected CSS variables

### Checklist


src/components/FeedbackMessage/FeedbackMessage.module.css
```diff
@@ -1,23 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.feedbackMessage {
```
fredmaggiowski, line 19: There's a typo here 😭

src/components/ThemeProvider/utils/themeToVariables.ts
```diff
@@ -18,6 +18,14 @@
 
 import { isObject } from 'lodash-es'
 
+const varsInPixelSegments = ['shape', 'spacing', 'fontSize', 'lineHeight']
+
+function addUnitToValue(value: unknown, fieldName: string): unknown {
+  return varsInPixelSegments.some((segment) => fieldName.includes(segment))
```
fredmaggiowski, line 24: Why includes here instead of a stricter comparison?

fedepini (approved):

src/components/ThemeProvider/utils/themeToVariables.ts
epessina: Since the names of the variables are not all the same (e.g., all starting with one of these keywords) it seemed more flexible to just check with an `includes`. 

However, if needed, I can make an equality check listing all the variables of interest

fredmaggiowski: No actual need for it something stricter, not that it comes to my mind at least. If you think it's safer this way it's fine to me

Danielecina: it is important that designers generate the theme without px or be aligned on this change

FrankieBortot: Af of today, we have no control over the plugin we use to export the primitives, and it automatically generates the values with "px" attached.
Screenshot 2024-06-04 at 12 10 55

epessina: The change made to the library only regards the CSS variables, not the theme itself. It was my understanding that Ant needs the values to be unitless, so we cannot change that part of the generation.
Therefore, the changes should not affect the design part, or the theme generation

davidebianchi (approved):
LGTM

PR #494, opened 2024-06-14
feat: drawer component

fredmaggiowski:
### Description

This PR introduces the first implementation for the new Drawer component!! 🎉 

Some important notes on missing implementations:

* The header close button is not included in this PR: the Design System specification requires the X to be positioned on the right of the header, however antd positions it on the left and does not provide any means to change that. I decided not to take any further implementation unless we are sure about the design choice
* Unlike a modal footer, which always displays buttons, the Drawer footer may contain a completely customized set of components. This PR introduces the footer prop that accepts a react node for this second use-case, it can be evolved in the future to be `ReactNode | { buttons: ReactNode[] }`

### Checklist

- [X] tests are included


package.json
```diff
@@ -41,7 +41,7 @@
   },
   "dependencies": {
     "@fontsource/inter": "^5.0.18",
-    "antd": "5.9.2",
+    "antd": "5.13.0",
```
fredmaggiowski, line 44: I need this update for semantic footer styling  https://ant.design/components/drawer#semantic-dom

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { Drawer as AntdDrawer } from 'antd'
+import { ReactElement } from 'react'
+
+import { DrawerProps } from './Drawer.props'
+import { Footer } from './Drawer.Footer'
+import { Title } from './Drawer.Title'
+
+const styles = {
+  footer: { padding: '24px' },
```
fredmaggiowski, line 27: To prevent this hardcoded value I could use a css module that defines the class, however I'd then need to pass the styles prop to the antd drawer in order to reset the padding to 0.

The implementation would look like the following:


```css
.drawer {
  footer {
    padding: var(--spacing-padding-xl, 24px)
  }
}
```

```ts
import styles from './Drawer.module.css'

const { drawer } = styles

const antdDrawerStyles = { footer: { padding: 0 } }

....

   <AntdDrawer
      className={drawer}
      ....
      styles={antdDrawerStyles}
....
```

src/components/Drawer/Drawer.Title.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+
+export type DrawerTitle = ReactNode
+
+export type TitleProps = {
+  title: DrawerTitle,
+}
+
+export const Title = ({ title }: TitleProps): ReactElement => {
+  const ellipsis = useMemo(() => ({ rows: 1, tooltip: title }), [title])
```
davidebianchi, line 31: It's better to use memo in this case: https://react.dev/reference/react/memo. This element should be rerendered only when title changes

davidebianchi (approved):
LGTM

src/components/Drawer/Drawer.Title.tsx
fredmaggiowski: you mean by doing such as:

```ts
export const Title = memo(({ title }: TitleProps): ReactElement => {
   ...
})
```

Doing so may require changes in the future if we add more props other than title to render other stuff and need partial re-renders within the Title component 🤔

src/hooks/useDrawer/useDrawer.tsx
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+import { useCallback, useState } from 'react'
+
+import { DrawerAPI } from './types'
+
+/**
+ * useDrawer hook allows for hiding visibility state .
+ *
+ * @returns {DrawerAPI} An object which includes the Drawer component, the drawer state (visible or hidden) and methods
```
epessina, line 26: The returned object does not include the component. We should either change the doc or add a memorized component to the hook (as in the `useModal` hook). I would go for the latter, otherwise this hook seems a bit redundant

src/components/Drawer/Drawer.props.ts
```diff
@@ -0,0 +1,64 @@
...
+     */
+    isVisible?: boolean,
+
+    /**
+     * React drawer key
+     */
+    key?: string,
+
+    /**
+     * Function invoked at the click of the Drwaer close (X) button.
```
epessina, line 56:
```suggestion
     * Function invoked at the click of the Drawer close (X) button.
```

epessina (changes requested):

src/hooks/useDrawer/useDrawer.tsx
fredmaggiowski: >The returned object does not include the component. 

this is sadly intended, I've started developing the useDrawer hook as a mean of hiding the visible state **and** keeping it associated with a component (as the useModal does); however there is a problem with the visibility state management within the hook that causes the re-creation of the component and the lack of animations during the drawer close.

For this first implementation I decided not to include the component in the hook result as it would produce a buggy drawer, we should investigate it in the future to make the useDrawer hook more useful since, at least it still hides the state management part...

I'm changing the doc here

epessina (approved):

davidebianchi (approved):

PR #524, opened 2024-07-05
fix: table actions allows for action.onClick to control icon

fredmaggiowski:
### Checklist

- [ ] commit message and branch name follow conventions
- [ ] tests are included
- [ ] changes are accessible and documented from components stories
- [ ] you are not committing extraneous files or sensitive data
- [ ] necessary labels have been applied to this pull request (enhancement, bug, ecc.)

### DIFF
--- a/src/components/Table/Table.test.tsx
+++ b/src/components/Table/Table.test.tsx
@@ -18,6 +18,7 @@
 
 import { WithExternalFiltersAndSorters, alignedColumns, columns, customActions, data, expandable, filteredAndSortedColumns, footer, hugeData, pagination, rowKey, rowSelection, sizedColumns, spannedColumns } from './Table.mocks'
 import { fireEvent, render, screen, waitFor, within } from '../../test-utils'
+import { Action } from './Table.types'
 import { Table } from '.'
 
 describe('Table Component', () => {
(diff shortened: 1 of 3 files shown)


src/components/Table/Table.types.ts
```diff
@@ -69,7 +69,7 @@ export type CustomAction<RecordType extends GenericRecord> = {
   icon: ReactNode,
   isDanger?: boolean,
   isDisabled?: (record: RecordType, index?: number) => boolean | boolean,
-  onClick: (record: RecordType, index: number | undefined, event: React.MouseEvent<Element, MouseEvent>) => void,
+  onClick?: (record: RecordType, index: number | undefined, event: React.MouseEvent<Element, MouseEvent>) => void,
```
GiovannaMonti, line 72: does it make sense to have an action without the action logic?

src/components/Table/Table.test.tsx
```diff
@@ -215,6 +216,42 @@ describe('Table Component', () => {
...
+
+    const editButtons = screen.getAllByRole('button', { name: 'PiPencilSimpleLine' })
+    const deleteButtons = screen.getAllByRole('button', { name: 'PiTrash' })
+
+    expect(editButtons).toHaveLength(props.data.length)
+    expect(deleteButtons).toHaveLength(props.data.length)
+
+    const [editButton] = editButtons
+    expect(editButton).toBeVisible()
+    fireEvent.click(editButton)
```
GiovannaMonti, line 244: can we use `userEvent`?

src/components/Table/Table.test.tsx
```diff
@@ -215,6 +216,42 @@ describe('Table Component', () => {
...
+    expect(onEditRow).toHaveBeenCalledTimes(1)
+    expect(onEditRow).toHaveBeenCalledWith(props.data[0], 0, expect.any(Object))
+
+    const [, deleteButton] = deleteButtons
+    expect(deleteButton).toBeVisible()
+    fireEvent.click(deleteButton)
```
GiovannaMonti, line 250: can we use `userEvent`?

danielemarostica (approved):

src/components/Table/Table.types.ts
MIA-Deltat1995: yes, because edit and delete actions could be defined directly as table props, but i think that is better force to use onClick if some custom edit/delete actions are defined and so ignore the onEditRow/onDeleteRow props in this case.

fredmaggiowski: The on click priorityis exactly how you describe it

src/components/Table/Table.test.tsx
fredmaggiowski: I've used fireEvent to be coherent with the other tests in the file

PR #535, opened 2024-07-12
feat(button): Add block prop

LiliumCandidum:
### Description

This PR aims to add the `block` property to the button component. This property enables the button to fit the parent container witdh.

### Checklist

- [x] tests are included

### DIFF
--- a/src/components/Button/Button.props.ts
+++ b/src/components/Button/Button.props.ts
@@ -25,7 +25,7 @@ export type ButtonProps = {
   /**
    * Fits button width to its parent width
    */
-  block?: boolean,
+  isBlock?: boolean,
 
   /**
    * The children nodes to be rendered within the button context.
--- a/src/components/Button/Button.props.ts
+++ b/src/components/Button/Button.props.ts
@@ -23,6 +23,11 @@ import { HTMLType, Hierarchy, IconPosition, Shape, Size, Type } from './Button.t
 export type ButtonProps = {
 
   /**
+   * Fits button width to its parent width
+   */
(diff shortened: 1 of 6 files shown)


src/components/Button/Button.test.tsx
```diff
@@ -152,4 +152,9 @@ describe('Button Component', () => {
     const button = screen.getByTitle('title test')
     expect(button).toBeVisible()
   })
+
+  test('rendersblock button correctly', () => {
```
epessina, line 156:
```suggestion
  test('renders block button correctly', () => {
```

epessina (dismissed):

src/components/Button/Button.props.ts
```diff
@@ -22,6 +22,11 @@ import { HTMLType, Hierarchy, IconPosition, Shape, Size, Type } from './Button.t
 
 export type ButtonProps = {
 
+  /**
+   * Fits button width to its parent width
+   */
+  block?: boolean,
```
fredmaggiowski, line 28: boolean prop should use the format "isBlock", "canBlock" etc (e.g. isVisible instead of visible)

This rule has never been written yet, we should add it to the contributing guidelines

fredmaggiowski: could it be "mustFitParent" ?

LiliumCandidum: I think "isBlock" is fine

Danielecina (approved):

epessina (approved):

PR #536, opened 2024-07-12
feat(core): add data-testid to components

### LINKED ISSUE #372: `data-testid` not available in Design System components

While using the Design System to migrate part of the UI in our application, we noticed that sometimes we use the `data-testid` attribute for testing purposes. However, the components in the DS does not expect this attribute, therefore we have to change our tests as well.

Do you think that the `data-testid` attribute will be included in the components?

## A snippet of code for replicating the issue or showing the proposal usage if applicable
If I've a button component like the following:
```
<Button data-testid={'my-button-cancel'} onClick={onCancel} />
```
and a test (with Jest and React Testing Library) with this instruction:
```
userEvent.click(screen.getByTestId('my-button-cancel'))
```
The test will fail since the button won't have the `data-testid` atrtribute.

For the `Button` component, as example:
```
export const Button = ({
  "data-testid": dataTestid,
  ...
}: ButtonProps): ReactElement  => {
 ...
return (
    <AntButton
      ...
      data-testid={dataTestid}
      ...
    />
  )
}
```

### Your environment

node: v20.11.1

os: Ubuntu 22.04

LiliumCandidum:
### Description

This PR aims to add the `dataTestId` prop to almost every component.
I didn't add the prop to the components that I thought it was not necessary but it can be discussed and I'm not quite sure about the Icon component.

The components with the new prop are:
- Button
- Breadcrumb
- Drawer
- Menu
- SegmentedControl
- Modal
- Table
- FeedbackMessage

The components **without** the prop are:
- Divider
- Icon
- ThemeProvider
- Typography

### Addressed issue

Closes #372

### Checklist

- [x] tests are included

### DIFF
--- a/src/components/Drawer/Drawer.tsx
+++ b/src/components/Drawer/Drawer.tsx
@@ -30,6 +30,7 @@ const closeIcon = <Icon color="currentColor" name="PiX" size={16} />
 
 export const Drawer = ({
   children,
+  dataTestId,
   destroyOnClose,
   docLink,
   footer,
@@ -43,6 +44,7 @@ export const Drawer = ({
     <AntdDrawer
       className={styles.drawer}
       closeIcon={closeIcon}
+      data-testid={dataTestId}
       destroyOnClose={destroyOnClose}
       footer={footer && <Drawer.Footer footer={footer} />}
       id={id}
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -47,6 +47,7 @@ export const defaults = {
  * @returns {Menu} Menu component
  */
 export const Menu = ({
+  dataTestId,
   defaultOpenKeys = defaults.defaultOpenKeys,
   defaultSelectedKey,
   hierarchy = defaults.hierarchy,
@@ -82,6 +83,7 @@ export const Menu = ({
       >
         <AntMenu
           className={menuClassNames}
+          data-testid={dataTestId}
           defaultOpenKeys={defaultOpenKeys}
           defaultSelectedKeys={defaultSelectedKey ? [defaultSelectedKey] : undefined}
           // getPopupContainer is needed for nested menus to inherit CSS properties in the vertical mode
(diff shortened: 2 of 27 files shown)


src/components/Drawer/Drawer.tsx
```diff
@@ -43,6 +44,7 @@ export const Drawer = ({
     <AntdDrawer
       className={styles.drawer}
       closeIcon={closeIcon}
+      data-testid={dataTestId}
```
Danielecina, line 47: This prop does not exist in antd, so it does not work in theory. 
I suggest not use props if they are documented

src/components/Modal/Modal.test.tsx
```diff
@@ -149,4 +149,12 @@ describe('Modal Component', () => {
 
     expect(baseElement).toMatchSnapshot()
   })
+
+  test('should get modal by test id', async() => {
+    render(<Modal {...props} dataTestId="modal-test" />)
```
Danielecina, line 154: We cannot trust it to go if it is not documented by antd, because it could quietly break compatibility

Danielecina: Same for every component that use data-testid directly on antd

src/components/Menu/Menu.tsx
```diff
@@ -82,6 +83,7 @@ export const Menu = ({
       >
         <AntMenu
           className={menuClassNames}
+          data-testid={dataTestId}
```
Danielecina, line 86: idem here

src/components/Drawer/Drawer.tsx
davidebianchi: There are some tests, so if test passes it should work (and we can be notified if antd break this behaviour). We could trust on it?

Danielecina: why risk to do a DS breaking change if antd change the behaviour? 
I suggest to use a custom node or not use for this component the data-testid and use the traditional getByX to getting the node inside your test.

PR #537, opened 2024-07-12
feat(tag): added tag component

GiovannaMonti:
### Description

Added **Tag component** remapping only the fundamental props from the antd5 Tag.

### Checklist

- [x] tests are included

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,7 @@ You can see the `npm` scripts available inside the `package.json` file and use t
 :warning: All changes **MUST** pass through merge request approvals.
 
 ## Before submitting a PR
+
  * Are you in the contributors section inside the `package.json`? Add yourself!
 
 ## How to tag
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,6 @@ You can see the `npm` scripts available inside the `package.json` file and use t
 :warning: All changes **MUST** pass through merge request approvals.
 
 ## Before submitting a PR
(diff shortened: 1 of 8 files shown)


CONTRIBUTING.md
```diff
@@ -29,7 +29,7 @@ You can see the `npm` scripts available inside the `package.json` file and use t
 :warning: All changes **MUST** pass through merge request approvals.
 
 ## Before submitting a PR
-
+ * Your PR title should follow the `<type>(<scope>): <subject>` format, so that it gets correctly categorized in the release notes.
```
GiovannaMonti, line 32:
```suggestion
```

src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,47 @@
...
+  } : TagProps
+) : ReactElement => {
+  return (
+    <AntTag
+      bordered={isBordered}
+      closeIcon={closeIcon}
+      color={color}
+      onClose={onClose}
+    >
+      {children}
```
fredmaggiowski, line 44: I guess we are going to use it most of the time with our `FormattedMessage` component but a full ReactNode in the tag could be abused 🤔  would it make sense to enforce a Typography element or to enforce a prop definition compatbile with our typography?

Danielecina: I think it is important to ask the designer to know if the tag component will have in the future a specific size. But for i18n I suggest using a children's node

Danielecina (approved):

fredmaggiowski (approved):
LGTM

PR #547, opened 2024-07-22
Added type link to Button component

MIA-Deltat1995:
### Description

Added type link to button to obtain the anchor style removing padding and margin

##### <Changed component name>
    - Button

### Checklist

- [x] tests are included

### Button links on storybook
all

### Primary button link without margin and padding
primary

### Primary button link (default)
primary_default

### Primary button link (hover)
primary_hover

### Primary button link (active)
primary_active


src/components/Button/Button.props.ts
```diff
@@ -126,6 +126,7 @@ export type ButtonProps = {
    * - "outlined": medium-emphasis button with a stroke around the button container and no fill,
    *   typically used for secondary actions;
    * - "ghost": transparent button with no stroke, typically used for links.
```
fredmaggiowski, line 128: Please improve this doc line, why should ghost be typically used for links now that there's the link type

src/components/Button/Button.stories.tsx
```diff
@@ -101,6 +101,57 @@ export const DangerGhost: Story = {
   },
 }
 
+export const PrimaryLink: Story = {
```
fredmaggiowski, line 104: Is the code snippet from these stories correctly rendered?

fredmaggiowski (changes requested):

GiovannaMonti (dismissed):

src/components/Button/Button.props.ts
MIA-Deltat1995: Because there are cases where we want to preserve margin and padding of a button, but show the content as a link so without border and background color

src/components/Button/Button.stories.tsx
MIA-Deltat1995: I have inserted some screenshot on the PR description.

src/components/Button/Button.props.ts
fredmaggiowski: the doc still doesn't make sense, it suggests using this for links while there is a link type. if there are use cases for this to be used as link as well we have to describe them; otherwise let's remove the note.

src/components/Button/Button.stories.tsx
fredmaggiowski: not about the code snippets 😅

src/components/Button/Button.props.ts
MIA-Deltat1995: ok, it is better to remove the recommendation

src/components/Button/Button.stories.tsx
MIA-Deltat1995: Ah okok 

Schermata del 2024-07-23 10-30-17

fredmaggiowski (approved):
LGTM

PR #557, opened 2024-07-29
feat: update theme, HX components now have updated colors

ThisIsDemetrio:
### Description

There have been updates on the Theme including the following:

- components in typography (all the `HX`, `BodyX`, `action`, `link`, `caption`, `code`) now specify a color (`#636363`) (before: #464646 was used for `HX`, the other typography elements are the same);
- font family for `code` is now `IBM Plex Mono`, not `Menlo` (which is included as `code-alternative`);

The following components have been impacted by the color change:

- Card
- Drawer
- Modal

Please, check the snapshots in the proposed change for more information.

### Addressed issue

No issue raised for this particular feature.

### Checklist

- [X] tests are included


src/themes/files/dark/primitives.json
```diff
@@ -573,6 +573,10 @@
       "$value": "Inter"
     },
     "code": {
+      "$type": "string",
+      "$value": "IBM Plex Mono"
+    },
+    "code-alternate": {
```
GiovannaMonti, line 579: I don't know if it's possible, but maybe instead of creating an alternate var we could leverage the CSS functionality that allows you to list fonts in order of preference (e.g. putting "IBM Plex Mono, Menlo" as the value)

GiovannaMonti (approved):

src/themes/files/dark/primitives.json
ThisIsDemetrio: We can actually do that.
I'll tag @FrankieBortot to check it out with him too.

MIA-Deltat1995 (dismissed):

src/themes/files/dark/primitives.json
FrankieBortot: That's not really how tokens work: each token represent a clear and specific value, not a series of value.
Also, Figma doesn't allow that 🥲
I came up with "code-alternate" because 
1. it's a naming convention we already have
2. didn't want to change the original "code" token, so i hade little room for changes 
What would you suggest instead of "code-alternate"?

fredmaggiowski (dismissed):

FrankieBortot (approved):
LGTM

fredmaggiowski:
I'm not sure whether this is breaking or not 🤔

src/themes/files/dark/primitives.json
GiovannaMonti: Ok so it makes sense that the token should have only one value. 

My issue with the code-alternate here is the actual usefulness of the token: since that font is proprietary with a very strict license, and will hardly ever be used directly (I mean included in the project, with the license and all), the only use I see is as a fallback if the IBM one fails to be downloaded. The fallback applies only to Mac users, therefore in my opinion it's a bit too much to have it as a standalone value.

FrankieBortot: I agree, we can remove it from the primitives.
And we on Figma can have a separate collection for special "external" tokens: this way we don't pollute tokens with unwanted value, but at the same time we can have our primiteves/semantic separated from the rest if needed

src/components/Typography/HX/HX.module.css
```diff
@@ -0,0 +1,15 @@
...
+    color: var(--typography-h2-color, #636363);
+}
+
+.h3 {
+    color: var(--typography-h3-color, #636363);
+}
+
+.h4 {
+    color: var(--typography-h4-color, #636363);
+}
```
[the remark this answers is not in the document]
ThisIsDemetrio: Included

src/themes/files/dark/primitives.json
ThisIsDemetrio: Removed `code-alternate`

FrankieBortot (approved):

fredmaggiowski (approved):

PR #563, opened 2024-08-02
feat: Tree component

ThisIsDemetrio:
### Description

We add the **Tree** component, based on [Ant Design Tree component](https://ant.design/components/tree).

Includes most of the properties of the original Tree component, like: showing vertical lines for chidlren, showing icons before the node, checkbox on nodes, strict checking, callback on checking (by clicking the checkbox, if enabled) and/or on selecting item (by clicking on the name), fixed height and virtual scrolling.

Add a little tweak to ensure that, when showing lines, the icon to expand/collapse children is still a caret (and not the "+"/"-" icons).

Also:
- updated Tree tokens and Global tokens to be sure the right colors are shown based on the configured theme. Because of this, the `Palette` type has been expanded to include the `text` property
- updated the theme because of some updates on the Neutral primitives (for consistency on tree nodes hovering)

### Addressed issue

No particular issue opened for this component.

### Checklist

- [X] tests are included


src/components/Tree/Tree.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+  })
+
+  test('renders Tree', async() => {
+    const { asFragment } = render(<Tree treeData={defaultTreeData} />)
+
+    expect(asFragment()).toMatchSnapshot()
+
+    expect(screen.getByText('Vegetables')).toBeInTheDocument()
+    const [, vegetablesCaret] = screen.getAllByRole('img', { name: /caret-down/i })
+    fireEvent.click(vegetablesCaret)
```
Danielecina, line 96: I think you should use the "userEvent" library because you should apply some checks on user clicks. Fireevent is the lower API level

ThisIsDemetrio: Nothing against that, expect the warning showing up during the text execution:

```
When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
```

(btw, the `act` imported from `test-utils` folder results deprecated, the `act` included in `react` package does not solve the warning)

Should I go with `userEvent` anyway?

Danielecina: mmm, no. let's leave fireevent

Danielecina (approved):

GiovannaMonti (dismissed):
since we changed existing theme colors, is this breaking?

fredmaggiowski (approved):
LGTM

fredmaggiowski:
> since we changed existing theme colors, is this breaking?

the next tag will already include color changes, i'm labelling this as breaking as well

PR #577, opened 2024-08-12
feat(Dropdown): add `onOpenChange` and `getPopupContainer` props

malta895:
### Description

I'm working on a component that needs to be aware of the changes of the nested `Dropdown` open status. Therefore, I added the according props in order to expose the needed state to the outer components.

I also need to modify the menu container to fix some positioning issues

##### Dropdown

    - add `onOpenChange` prop, forwarded to the inner Antd component
    - add `getPopupContainer` prop, forwarded to the inner Antd component

### Checklist

- [X] tests are included

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -103,7 +103,7 @@ export const Dropdown = ({
       menu={menu}
       overlayClassName={classes}
       trigger={triggers}
-      onOpenChange={onOpenChange ? onOpenChangeInternal : undefined}
+      onOpenChange={onOpenChangeInternal}
     >
       {innerNode}
     </AntdDropdown>
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -20,7 +20,7 @@ import { Dropdown as AntdDropdown, type MenuProps as AntdMenuProps } from 'antd'
 import React, { ReactElement, ReactNode, useCallback, useMemo } from 'react'
 import classNames from 'classnames'
 
-import { DropdownClickEvent, DropdownItem, DropdownProps, DropdownTrigger, ItemLayout } from './props'
+import { DropdownClickEvent, DropdownItem, DropdownProps, DropdownTrigger, ItemLayout, OpenChangeInfoSource } from './props'
 import Label from './components/Label'
 import styles from './dropdown.module.css'
 
@@ -49,9 +49,8 @@ export const Dropdown = ({
   items,
   onClick,
   triggers = defaults.trigger,
-  open,
   onOpenChange,
-  placement,
+  getPopupContainer,
 }: DropdownProps): ReactElement => {
   const uniqueClassName = useMemo(() => `dropdown-${crypto.randomUUID()}`, [])
 
@@ -79,17 +78,35 @@ export const Dropdown = ({
 
   const classes = useMemo(() => classNames(styles.dropdownWrapper, uniqueClassName), [uniqueClassName])
 
+  const onOpenChangeInternal = useCallback(
+    (open: boolean, info?: {source: 'trigger'| 'menu'}) => {
+      if (!onOpenChange) {
+        return
+      }
+      switch (info?.source) {
+      case 'trigger':
+        onOpenChange(open, { source: OpenChangeInfoSource.Trigger })
+        break
+      case 'menu':
+        onOpenChange(open, { source: OpenChangeInfoSource.Menu })
+        break
+      default:
+        onOpenChange(open)
+      }
+    },
+    [onOpenChange]
+  )
+
   return (
     <AntdDropdown
       autoFocus={autoFocus}
       disabled={isDisabled}
       dropdownRender={dropdownRender}
+      getPopupContainer={getPopupContainer}
       menu={menu}
-      open={open}
       overlayClassName={classes}
-      placement={placement}
       trigger={triggers}
-      onOpenChange={onOpenChange}
+      onOpenChange={onOpenChangeInternal}
     >
       {innerNode}
     </AntdDropdown>
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -49,6 +49,8 @@ export const Dropdown = ({
   items,
   onClick,
   triggers = defaults.trigger,
+  open,
+  onOpenChange,
 }: DropdownProps): ReactElement => {
   const uniqueClassName = useMemo(() => `dropdown-${crypto.randomUUID()}`, [])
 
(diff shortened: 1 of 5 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -82,8 +84,10 @@ export const Dropdown = ({
       disabled={isDisabled}
       dropdownRender={dropdownRender}
       menu={menu}
+      open={open}
```
fredmaggiowski, line 87: Is the open prop really needed? Do you need to control visibility from the outside?

src/components/Dropdown/props.ts
```diff
@@ -118,4 +118,19 @@ export type DropdownProps = {
...
+
+  /**
+   * Called when the open state changes. Not triggered when hidden by click item.
+   *
+   * @param open the current open state of the Dropdown
+   * @param info an object with the following props:
+   * - source: the source of the open change, can be 'trigger' or 'menu'
+   * @returns
+   */
+    onOpenChange?: (open: boolean, info: {source: 'trigger' | 'menu'}) => void
```
fredmaggiowski, line 135: define an enum for mapping the values

fredmaggiowski (changes requested):

src/components/Dropdown/Dropdown.tsx
malta895: Not as of now, but I thought it made sense to have it together with `onOpenChange`, plus its documentation refers to this prop. 
Do you think we should wait to add it until we actually need it?

fredmaggiowski: onOpenChange does not directly refers to the open prop, more on the current open state which is managed by the component itself. Open would allow you to control the dropdown visibility from the outside. While it makes sense to provide such feature I'd rather discourage its use tbh.

src/components/Dropdown/props.ts
malta895: It would make it incompatible with the underlying type of Antd, which is defined exactly as I did here

malta895: I could define it with a type instead? Such as:
```typescript
export type OpenChangeInfoSource = 'trigger' | 'menu'
```
This would keep the compatibility

fredmaggiowski: better using enum for developer experience when using the component. I'd expect to use something like `Dropdown.OpenChangeSource.Trigger` and `Dropdown.Placement.Top`

src/components/Dropdown/Dropdown.tsx
malta895: Ok I get your point, I removed it and left only the `onOpenChange`, improving its documentation

src/components/Dropdown/props.ts
malta895: defined an enum and a function to remap it to the underlying antd function correctly

danielemarostica (dismissed):

src/components/Dropdown/Dropdown.tsx
```diff
@@ -76,14 +78,35 @@ export const Dropdown = ({
 
   const classes = useMemo(() => classNames(styles.dropdownWrapper, uniqueClassName), [uniqueClassName])
 
+  const onOpenChangeInternal = useCallback(
+    (open: boolean, info?: {source: 'trigger'| 'menu'}) => {
+      if (!onOpenChange) {
+        return
+      }
+      switch (info?.source) {
```
fredmaggiowski, line 86: I'd simplify this code section and make it so the `onOpenChange` gets called only once.

The switch case could be substituted by a simple map, and something like this should do the trick:

```typescript
// outside the component
type antdTriggerSource = 'trigger'|'menu'
const antdSourceMap: Record<antdTriggerSource,OpenChangeInfoSource> = {
'trigger': OpenChangeInfoSource.Trigger,
'menu': OpenChangeInfoSource.Menu,
}

// here
const source = antdSourceMap[info?.source]
onOpenChange(open, source ? {source} : undefined)
```

src/components/Dropdown/props.ts
```diff
@@ -80,6 +80,15 @@ export type DropdownClickEvent = {
   selectedPath: string[],
 }
 
+export enum OpenChangeInfoSource {
+  Trigger='trigger',
```
fredmaggiowski, line 84: please use consistent spacing with other enums

```suggestion
  Trigger = 'trigger',
```

src/components/Dropdown/props.ts
```diff
@@ -118,4 +127,24 @@ export type DropdownProps = {
...
+   *    (e.g. the dropdown is opening because the children was clicked, or closing because the focus was lost)
+   * - menu: the open state was changed by clicking on a menu item
+   *    (e.g. the dropdown is closing because a menu item was clicked)
+   * @returns
+   */
+   onOpenChange?: (open: boolean, info?: OpenChangeInfo) => void
+
+   /**
+    * To set the container of the dropdown menu.
+    * The default is to create a div element in body,
```
fredmaggiowski, line 146: I didn't get this sentence, you mean that by default the dropdown div will be appended in the body?

fredmaggiowski (changes requested):

src/components/Dropdown/Dropdown.tsx
malta895: `antdSourceMap[info?.source]` leads to a type error due to the fact that `source` cannot be `undefined`, and neither the object key can be.
I had to leave the check at runtime, still simplifying a little bit the code

src/components/Dropdown/props.ts
malta895: fixed

src/components/Dropdown/props.ts
malta895: That's what Antd documentation says.
 By inspecting the element, I actually observed the default behavior is to actually append it to the body elements, i.e. at the root level right before the `</body>` closing tag.

src/components/Dropdown/Dropdown.tsx
malta895: I actually managed to remove the check, it should be acceptable now

src/components/Dropdown/props.ts
fredmaggiowski: ok, I was more worried about the phrasing clearness tbh

src/components/Dropdown/Dropdown.tsx
```diff
@@ -76,14 +84,26 @@ export const Dropdown = ({
...
   return (
     <AntdDropdown
       autoFocus={autoFocus}
       disabled={isDisabled}
       dropdownRender={dropdownRender}
+      getPopupContainer={getPopupContainer}
       menu={menu}
       overlayClassName={classes}
       trigger={triggers}
+      onOpenChange={onOpenChange ? onOpenChangeInternal : undefined}
```
fredmaggiowski, line 106: since the `onOpenChange` is already verified at lin 89, you can simply leave here our internal callback set

```suggestion
      onOpenChange={onOpenChangeInternal}
```

fredmaggiowski (changes requested):

fredmaggiowski (approved):
LGTM

fredmaggiowski (approved):

PR #579, opened 2024-08-14
feat(dropdown): highlight selected items with initialSelectedItems and multiple highlight support

fredmaggiowski:
### Description

This PR introduces:

 - `Dropdown#initialSelectedItems` **optional** prop that helps selecting one or more items at first render.
 - `Dropdown#multiple` **optional** prop that can be used to control whether dropdown highlights should consider multiple items or a single one.

**Screenshot below shows highlight in multiple mode**

Screenshot 2024-08-14 alle 16 06 58

### Checklist

- [x] tests are included

### DIFF
--- a/src/components/Dropdown/Dropdown.test.tsx
+++ b/src/components/Dropdown/Dropdown.test.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable max-nested-callbacks */
 /**
  * Copyright 2024 Mia srl
  *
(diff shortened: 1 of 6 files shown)


src/components/Dropdown/props.ts
```diff
@@ -123,6 +123,8 @@ export type DropdownProps = {
    */
   onClick: (event: DropdownClickEvent) => void
 
+  selectedItems?: string[],
```
fredmaggiowski, line 126: doc!

src/components/Dropdown/Dropdown.test.tsx
```diff
@@ -233,10 +234,53 @@ describe('Dropdown Component', () => {
       expect(onOpenChange).toHaveBeenNthCalledWith(2, false, { source: OpenChangeInfoSource.Menu })
     })
   })
+
+  describe('highlights selectedItems', () => {
+    describe('single mode', () => {
```
danielemarostica, line 239: ha senso anche in multi?

fredmaggiowski: it does, indeed...

fredmaggiowski:
todo: bold primary label!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -77,10 +90,11 @@ export const Dropdown = ({
 
   const menu = useMemo(() => ({
     items: antdItems,
-    onClick: onAntdMenuClick,
     /* istanbul ignore next */
     getPopupContainer: (triggerNode: HTMLElement) => (document.querySelector(`.${uniqueClassName}`) || triggerNode) as HTMLElement,
-  }), [antdItems, onAntdMenuClick, uniqueClassName])
+    onClick: onAntdMenuClick,
+    selectedKeys: selectedItems,
```
malta895, line 96: does it make sense to check here or somewhere else whether multiple `selectedItems` have been passed with `multiple: false`, and in case only take the first or throw an error?

fredmaggiowski: there is no need for doing that here, we could make such assertion at first render; however what would be the point? we could:

- throw an error: dangerous at runtime
- ignore the provided `intialSelectedItems`: it would end up in a weird interface
- log a console.warn|error: possibly the best choice, however we have no usecase yet of such behaviours
- do nothing and accept and undefined behaviour (which is the strategy applied here)

At the end of the day is a simple misconfiguration, I'd go with the the last option.

malta895: Got the point, yeah I agree with you we can just ignore it

PR #593, opened 2024-09-03
fixed fullscreen modal regression with antd 5.13->5.17 update

nicmell:
The update of the Ant Design library from version 5.13 to 5.17 in [this commit](https://github.com/mia-platform/design-system/commit/016533a47b8a97f6ac38f3fc10a6ba6ed669cf28#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L50) has secretly changed the implementation of the Ant modal by adding an extra wrapper around the ant-modal-content element. The CSS fix present here restores the correct functionality of the fullscreen modal.

Screenshot 2024-09-03 alle 12 03 00

### DIFF
--- a/src/components/Modal/Modal.module.css
+++ b/src/components/Modal/Modal.module.css
@@ -186,6 +186,11 @@
   height: 100% !important;
   max-width: 100%;
 
+  > div {
+    width: 100%;
+    height: 100%;
+  }
+
   :global(.mia-platform-modal-content) {
     border-radius: 0;
   }
--- a/src/components/Modal/Modal.module.css
+++ b/src/components/Modal/Modal.module.css
@@ -186,13 +186,9 @@
   height: 100% !important;
   max-width: 100%;
 
-  > div {
-    width: 100%;
-    height: 100%;
-  }
-
   :global(.mia-platform-modal-content) {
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.module.css
```diff
@@ -186,6 +186,11 @@
   height: 100% !important;
   max-width: 100%;
 
+  > div {
```
fredmaggiowski, line 189: Can we give this div a classname (or target a more specific classname)

previous implementation was fragile and was broken with the antd update; this fix sound as fragile as the previous implementation to me :(

nicmell: You are right, this implementation is fragile. 
Fortunately doing like [here](https://github.com/mia-platform/design-system/pull/593/commits/8abb59a62080dfda99e21de3e7fdf212faed3e28#diff-e7cfca382e4c844d9e6190d694684e0d98ff1f4f03ee5faee86d283036e87af1R191) fixes the issue, and gives a cleaner solution as well

fredmaggiowski (commented):
LGTM

fredmaggiowski (approved):

PR #620, opened 2024-09-19
feat(Dropdown): including "selectable" prop, defaults to true

ThisIsDemetrio:
#### Dropdown

Including new prop, called **selectable**, if not included it defaults to `true`. 

When set to `false`, any selection will not be saved. This might be useful in case a dropdown is used as a selection menu (and we do not need to show any previous selection when re-opening the menu).

Includes tests and storybook.

### [IMPORTANT] PR Checklist

#### PR conventions

#### Additional code checks

- [X] Changes are covered by tests 
- [X] Changes to components are accessible and documented in the Storybook
- [X] The browser console does not contain errors

### DIFF
--- a/src/components/Dropdown/Dropdown.test.tsx
+++ b/src/components/Dropdown/Dropdown.test.tsx
@@ -275,6 +275,46 @@ describe('Dropdown Component', () => {
         expect(firstUpdated).toMatchSnapshot('after click render Label 1 is not selected')
         expect(secondUpdated).toMatchSnapshot('after click render Label 2 is selected')
       })
+
+      it('does not highlight elements if selectable is set to false', async() => {
+        const onClick = jest.fn()
+        const props: DropdownProps = {
+          ...defaultProps,
+          onClick,
+          selectable: false,
+        }
+
+        renderDropdown({ props })
+        const button = screen.getByText('test-trigger-button')
+        userEvent.click(button)
+        const el = await screen.findByRole('menuitem', { name: 'Label 1' })
+        expect(el).toMatchSnapshot('at first render Label 1 is not selected')
+
+        userEvent.click(el)
+
+        // eslint-disable-next-line max-nested-callbacks
+        await waitFor(() => expect(onClick).toHaveBeenCalled())
+
+        userEvent.click(button)
+        const firstUpdated = await screen.findByRole('menuitem', { name: 'Label 1' })
+        expect(firstUpdated).toMatchSnapshot('after click render Label 1 is still not selected')
+      })
+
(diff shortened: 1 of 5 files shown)


ThisIsDemetrio:
Additional updates can come soon. I just want to review the logic inside the `Dropdown` component to simplify the `selectedItems` state in case `selectable` is `false`.

src/components/Dropdown/Dropdown.test.tsx
```diff
@@ -275,6 +275,46 @@ describe('Dropdown Component', () => {
...
+
+        // eslint-disable-next-line max-nested-callbacks
+        await waitFor(() => expect(onClick).toHaveBeenCalled())
+
+        userEvent.click(button)
+        const firstUpdated = await screen.findByRole('menuitem', { name: 'Label 1' })
+        expect(firstUpdated).toMatchSnapshot('after click render Label 1 is still not selected')
+      })
+
+      it('render and ignores "initialSelectedItems" prop if "selectable" is true', async() => {
```
GiovannaMonti, line 303: why does it ignore the initialSelected value, if selectable is true?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -60,6 +60,7 @@ export const Dropdown = ({
   getPopupContainer,
   initialSelectedItems = defaults.initialSelectedItems,
   multiple,
+  selectable = true,
```
fredmaggiowski, line 63: what is the purpose of this "selectable" prop? I don't recall a dropdown where items are not selectable

GiovannaMonti: this applies to dropdowns that for example open another element on element click, so the item should not stay selected / highlighted

fredmaggiowski: ok clear, let's change the name since it's a bool,  like "shouldHighlightSelection" or something similar

src/components/Dropdown/Dropdown.test.tsx
ThisIsDemetrio: There was a mistake on the test. Should be fixed now.

GiovannaMonti (dismissed):

src/components/Dropdown/Dropdown.tsx
ThisIsDemetrio: Fair enough to consider a verb-like word since it is a bool. I'd suggest `persistSelection` as the prop not only would allow the highlight of previous choices but it works also on persisting values previously selected. Let me know if you like that.

fredmaggiowski (approved):
LGTM

PR #621, opened 2024-09-20
feat(Button): added download prop

nicmell:
Added download prop to indicate the href attribute refers to a file to be downloaded rather than displayed in the browser

### DIFF
--- a/src/components/Button/Button.props.ts
+++ b/src/components/Button/Button.props.ts
@@ -132,7 +132,7 @@ export type ButtonProps = {
 
   /**
    * Indicates that the linked resource is intended to be downloaded rather than displayed in the browser.
-   * If is a string, the values specifies the default file name for use in labeling the resource in a local file system.
+   * If it is a string, its value specifies the file name for use in labeling the resource in the local file system.
    */
   download?: string | boolean
 }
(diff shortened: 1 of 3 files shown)


src/components/Button/Button.props.ts
```diff
@@ -129,4 +129,10 @@ export type ButtonProps = {
    * - "link": button link without padding and margin;
    */
   type?: Type,
+
+  /**
+   * Indicates that the linked resource is intended to be downloaded rather than displayed in the browser.
+   * If is a string, the values specifies the default file name for use in labeling the resource in a local file system.
```
malta895, line 135:
```suggestion
   * If it is a string, the values specifies the default file name for use in labeling the resource in a local file system.
```

fredmaggiowski:
How does this interact with the onClick prop? can you add at least on story to showcase the feature and how to use it?

nicmell:
@fredmaggiowski I've gust added a story for the download prop.
The behaviour when both the onClick and download prop are set is:

the onClick callback is run, and the file referenced in the href is downloaded, unless the preventDefault() method on clickEvent passed as parameter to the callback is invoked

src/components/Button/Button.props.ts
nicmell: fixed in [[d1e589e](https://github.com/mia-platform/design-system/pull/621/commits/d1e589e2a7c23e659ef08dbecb80a9092dbc8701)](https://github.com/mia-platform/design-system/pull/621/commits/d1e589e2a7c23e659ef08dbecb80a9092dbc8701)

fredmaggiowski (approved):
LGTM

PR #629, opened 2024-09-24
fix(Button): changed design token's schema for neutral buttons

nicmell:
changed design token's schema for neutral buttons


src/themes/files/dark/theme.json
```diff
@@ -189,11 +189,10 @@
         "contrastText": "#ffffff"
       },
       "secondary": {
-        "active": "#636363",
-        "disabled": "#e5e5e5",
-        "focus": "#636363",
-        "hover": "#898989",
-        "selected": "#464646"
+        "main": "#ffffff",
```
fredmaggiowski, line 192: this sounds like a breaking change, isn't it?

nicmell: Well, yes if someone is using the old variables directly. I've checked in the design-system code and they aren'tused directly anywhere. but i do agree in mentioning it as a breaking change in the design token structure, so i'll apply the label and mentioning it in the changelog

Issue #530, opened 2024-07-10
Add new componente: Dropdown


FrankieBortot:
Proposal for new Dropdown component, to be used when interacting with inputs such as Select and Multiselect.

You can find the documentation :
- directly [on Figma](https://www.figma.com/design/0ar6alIEDe8iYCb8kU7Rxd/%5BDS%5D-Console-Mia-Platform?node-id=15416-32190&t=2MMibzVRJfQdxJMp-4), where the component lies
- [on Confluence](https://makeitapp.atlassian.net/wiki/spaces/MP/pages/4351164417/Guida+alla+dropdown)

GiovannaMonti:
BRUH

fredmaggiowski:
Implementation checklist:

* [ ]: basic implementation of items with desired formatting
* [ ]: cascading options
* [ ]: customizable heading and footer
* [ ]: search
* [ ]: icon support

PR #635, opened 2024-10-03
feat(Dropdown): add footer with custom actions

fredmaggiowski:
### Description

Added dropdown footer with customizable actions

immagine

### Addressed issue

#530 

### [IMPORTANT] PR Checklist

#### PR conventions

#### Additional code checks

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -104,7 +104,7 @@ export const Dropdown = ({
   const footerActionHook = useCallback(() => {
     const el = document.querySelector(`.${randomClass}`)
     // This branch is not testable since the dropdown always exists when the dropdown is
-    /* istanbul ignore next */
+    /* istanbul ignore if */
     if (!el) {
       return
     }
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -94,11 +94,14 @@ export const Dropdown = ({
   )
 
   const dropdownRender = useCallback((menu: ReactNode): ReactNode => {
+    if (!footer) {
+      return React.cloneElement(menu as ReactElement)
+    }
     return (
-      <>
+      <div className={styles.dropdownRenderContainer}>
         {React.cloneElement(menu as ReactElement)}
         <Footer footer={footer} />
-      </>)
+      </div>)
   }, [footer])
 
   const menu = useMemo(() => ({
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -123,7 +123,7 @@ export const Dropdown = ({
     }
 
     return (
-      <div className={styles.dropdownRenderContainer}>
+      <div className={styles.dropdownWithFooterRenderWrapper}>
         {React.cloneElement(menu as ReactElement, { style: { boxShadow: 'none' } })}
         <Divider margin={spacing?.margin?.none} />
         <Footer footer={hookedFooter} />
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -21,9 +21,11 @@ import React, { ReactElement, ReactNode, useCallback, useMemo, useState } from '
 import classNames from 'classnames'
 
 import { DropdownClickEvent, DropdownItem, DropdownProps, DropdownTrigger, ItemLayout, OpenChangeInfoSource } from './props'
-import { Footer } from './components/Footer'
+import { Footer, useFooterWithHookedActions } from './components/Footer'
+import { Divider } from '../Divider'
 import Label from './components/Label'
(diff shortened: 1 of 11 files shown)


src/components/Dropdown/components/Footer/index.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+describe('useFooterWithHookedActions', () => {
+  const hook = jest.fn()
+  const footer: DropdownFooter = {
+    top: 'top',
+    bottom: 'bottom',
+    actions: [
+      { onClick: jest.fn() },
+      { onClick: jest.fn() }],
+  }
```
Danielecina, line 35:
```suggestion
      { onClick: jest.fn() }
    ],
  }
```

src/components/Dropdown/dropdown.module.css
```diff
@@ -74,6 +74,13 @@
   }
 }
 
+.dropdownRenderContainer {
```
Danielecina, line 77: Is it possible to use design tokens?

src/components/Dropdown/props.ts
```diff
@@ -63,6 +63,45 @@ export enum DropdownTrigger {
...
+   * */
+  label?: ReactNode
+
+  /**
+   * callback invoked on click.
+   */
+  onClick: () => void
+}
+
+export type DropdownFooter = {
```
Danielecina, line 87: Hmm... Are you sure you are so strict about the footer?
Why not just use a simple react node? 
If I need another type of node that is not a button in the future? 
I am scared to be really so specific

src/components/Dropdown/Dropdown.tsx
```diff
@@ -91,9 +99,37 @@ export const Dropdown = ({
...
 
+  const randomClass = useMemo(() => crypto.randomUUID(), [])
+
+  /**
+   * This function is used to forcibly close the dropdown without controlling
+   * the `open` state via prop.
+   */
+  const footerActionHook = useCallback(() => {
+    const el = document.querySelector(`.${randomClass}`)
+    // This branch is not testable since the dropdown always exists when the dropdown is
```
malta895, line 110: is this comment incomplete?

src/components/Dropdown/props.ts
fredmaggiowski: > Why not just use a simple react node?

We discussed it yesteday. If the custom node introduces some buttons or other interactive elements the dropdown will not closed.

src/components/Dropdown/dropdown.module.css
fredmaggiowski: nope, [available tokens](https://ant.design/components/dropdown#design-token) are limited to `paddingBlock` and `zIndexPopup`

fredmaggiowski: Btw: top/bottom props may be exploited to introduce custom footer nodes. However I'd not provide such a freedom of use since it would degrade UX

src/components/Dropdown/props.ts
fredmaggiowski: Btw: top/bottom props may be exploited to introduce custom footer nodes. However I'd not explicitly provide such a freedom of use since it would degrade UX

src/components/Dropdown/dropdown.module.css
Danielecina: You can override the generic to specific component

```
{
  "components": {
    "Table": {
      "colorBgContainer": "#456578"
    }
  },
  ...
}
```

src/components/Dropdown/props.ts
Danielecina: > We discussed it yesterday. If the custom node introduces some buttons or other interactive elements the dropdown will not close.

I mean for the footer to be a generic react node and not such a specific interface. Let's have more freedom to pass in the footer anything no? 
The user is responsible for what they will put in the footer. So let's pass a structure from the outside.

src/components/Dropdown/dropdown.module.css
fredmaggiowski: I've verified and while the approach would be correct it would not work as this class is used on a custom component that acts as the original antd wrapper, so updating the theme doesn't act on my specific component.

I'm renaming the class to make this more explicit!

src/components/Dropdown/props.ts
fredmaggiowski: as said above: if there are interactive items elements in such footer we would have to deal either with:

- terrible UX: the dropdown doesn't close on footer button lick
- terrible DX: we would have to implement each time the open state management

The solution proposed here provides a structured way to define actions (that translates to button, that could be customizable in the future with new layout-specific props) and a set of `top`/`bottom` props to specify custom nodes.

If you need a footer with a non-interactive elements you can safely use `top` and provide any custom node. If you want to use an interactive elements you need to use actions so that we can take care of the UX

Danielecina (approved):

PR #674, opened 2024-10-25
feat(Modal): scrollbar on content instead of body, and better padding to not render scrollbars attached to content

ThisIsDemetrio:
##### Dropdown

From now on:

- the scrollbar is not on the `body` but on the `content` class (at the same level of the `aside` element)
- updated margins on the right side of the 'content` and the left side of the `aside` to avoid the effect that the scrollbars are too close to the content of the `content` element.
- 
### [IMPORTANT] PR Checklist

- [X] I am aware of standards and conventions adopted in this repository, defined in the [CONTRIBUTING.md file](https://github.com/mia-platform/design-system/blob/main/CONTRIBUTING.md)

#### PR conventions

- [X] PR title follows the `<type>(<scope>): <subject>` structure
- [X] The PR has been labeled according to the type of changes (e.g., enhancement, bug).

#### Additional code checks

- [X] Changes are covered by tests 
- [X] Changes to components are accessible and documented in the Storybook
- [X] The browser console does not contain errors


src/components/Modal/Modal.module.css
```diff
@@ -96,30 +94,38 @@
     }
 
     .content {
-      width: 100%;
+      min-width: 0;
```
Danielecina, line 97: Do you think it's important to know why this part was included?
_(avoid content or aside overflowing)_

Danielecina (dismissed):

src/components/Modal/Modal.module.css
ThisIsDemetrio: Sure. Let me know if the comment added is ok.

fredmaggiowski: reading it made it clear to me without previous knowledge, so I guess it's fine :)

fredmaggiowski (approved):
LGTM

src/components/Modal/Modal.module.css
```diff
@@ -96,30 +94,39 @@
     }
 
     .content {
-      width: 100%;
+      /* min-width to prevent both this element (.content) and sibling element (.asideWrapper) to overflow their continers */
+      min-width: 0;
       height: 100%;
       display: flex;
       flex-direction: column;
+      padding: 0 var(--spacing-padding-sm, 8px) 0 var(--spacing-padding-xl, 24px);
```
ThisIsDemetrio, line 102: This could represent a problem.

In case the modal is opened without the aside panel, the only padding the modal will have is this one, causing the right padding to be only of 8px and the left padding to be of 24px. This could cause an effect where the content of the modal stretch to the right side far more than the right side.

Danielecina (approved):

PR #681, opened 2024-10-30
feat(RadioGroup): Added RadioGroup component

guidozoli:
### RadioGroup
Added RadioGroup component

image

### [IMPORTANT] PR Checklist

#### PR conventions

#### Additional code checks

- [x] New components are exported from the `index` file


src/components/GroupRadio/GroupRadio.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { Radio } from './components/Radio'
+import styles from './GroupRadio.module.css'
+
+const { options: optionsClass, groupRadio } = styles
+
+const defaults = {
+  disabled: false,
+}
+
+export const GroupRadio = ({
```
fredmaggiowski, line 32: Isn't the more commonly accepted name "RadioGroup" ? 🤔

guidozoli: You're right, I renamed it to "RadioGroup"

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,69 @@
...
+  defaultValue,
+  disabled = defaults.disabled,
+  options,
+  onChange,
+}: RadioGroupProps): ReactElement => {
+  const [value, setValue] = useState<string | number | boolean >(defaultValue)
+
+  const radioOptions = useMemo((): ReactNode => {
+    return options.map((option) => (
+      <div key={option.value.toString()}>
```
Danielecina, line 42: why can't use the key directly to the Radio element?

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,69 @@
...
+  disabled: false,
+}
+
+export const RadioGroup = ({
+  defaultValue,
+  disabled = defaults.disabled,
+  options,
+  onChange,
+}: RadioGroupProps): ReactElement => {
+  const [value, setValue] = useState<string | number | boolean >(defaultValue)
```
Danielecina, line 38: Why can't RadioGroup directly use the defaultValue prop?

src/components/RadioGroup/RadioGroup.module.css
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.options {
+  display: flex;
+  flex-direction: column;
+  gap: var(--spacing-gap-lg, 16px)
+}
```
fredmaggiowski, line 23: new line at EOF

```suggestion
}

```

src/components/RadioGroup/components/Radio/index.module.css
```diff
@@ -0,0 +1,63 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.radio {
```
Danielecina, line 19: Didn't Antd component design tokens help to avoid this?

src/components/RadioGroup/RadioGroup.props.ts
fredmaggiowski, line 1: avoid having all files starting with the componente name, which is already expressed in the directory name. Let's follow the dropdown directory structure for an example

src/components/RadioGroup/RadioGroup.props.ts
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { RadioChangeEvent } from 'antd'
+
+export type RadioGroupOption = {
+  value: string | number | boolean;
```
fredmaggiowski, line 22: would it make sense to use a generic T to allow any kind of value to be supplied?

src/components/RadioGroup/RadioGroup.tsx
fredmaggiowski: I agree, keeping the defaultValue updated onchange sounds like a bit of overhead, is it actually required by antd to work it like this?

src/components/RadioGroup/components/Radio/index.module.css
Danielecina: I think not

src/components/RadioGroup/RadioGroup.tsx
fredmaggiowski: if value becomes generic it would be dangerous to use `toString`, I guess this is one of those circumstances where the map index would be more suitable as key than something so generic 🤔

src/components/RadioGroup/components/Radio/index.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  it('renders correctly with label and value props', () => {
+    const { asFragment } = render(<Radio {...baseProps} />)
+    expect(asFragment()).toMatchSnapshot()
+  })
+
+  it('renders correctly with label, value and description props', () => {
+    const { asFragment } = render(
+      <Radio {...{ ...baseProps, description: 'description test' }} />
+    )
```
Danielecina, line 42:
```suggestion
    const { asFragment } = render(<Radio { ...baseProps, description: 'description test' } />)
```

src/components/RadioGroup/RadioGroup.tsx
fredmaggiowski, line 42: if value becomes generic it would be dangerous to use toString, I guess this is one of those circumstances where the map index would be more suitable as key than something so generic 🤔

src/components/RadioGroup/components/Radio/index.props.ts
fredmaggiowski, line 1: why are there this many prop files? I'd define all props in a single one

src/components/RadioGroup/index.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export { RadioGroup } from './RadioGroup'
```
fredmaggiowski, line 19: The RadioGroup component must be exported from the main index as well

fredmaggiowski (changes requested):
For sure exporting the index and prop renaming is a must, I'd consider also the generic value (if that's compatible with antd interface)

src/components/RadioGroup/components/Radio/index.theme.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Generates a Ant theme configuration for Radop component based on a theme configuration.
+ *
+ * @link https://ant.design/components/Radop#design-token
+ *
+ * @param {Partial<Theme>} theme - theme configuration.
+ * @returns {Partial<ComponentsTheme>} The generated Input Ant theme configuration.
+ */
+export default ({ palette, spacing }: Partial<Theme>): ComponentsTheme['Radio'] => ({
+  dotSize: Number(spacing?.padding?.sm),
```
Danielecina, line 31: This spacing?.padding?.sm type is a generic number, no? Do we want it to be 8 fixed?

src/components/RadioGroup/RadioGroup.tsx
guidozoli: the linter complains for the rule "react/no-array-index-key". Should I ignore it for this line?

Danielecina: > il linter si lamenta per la regola "react/no-array-index-key". Dovrei ignorarla per questa riga?

nope. can we use key directly to `<Radio key={} />` istead to a wrapper div?

src/components/RadioGroup/RadioGroup.props.ts
guidozoli: I changed in order to have a generic value type

src/components/RadioGroup/RadioGroup.props.ts
```diff
@@ -0,0 +1,40 @@
...
+  label: string;
+  description?: string;
+  disabled?: boolean;
```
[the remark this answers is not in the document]
guidozoli: fixed

src/components/RadioGroup/RadioGroup.tsx
guidozoli: I thought in Ant examples always used the value state, but actually in the example with the default value is not used. 
I passed the devfaultValue prop directly to Radio.Group and removed the value state, and everything seems to work fine

src/components/RadioGroup/RadioGroup.tsx
guidozoli: I removed the wrapping div and used key directly in Radio element

src/components/RadioGroup/index.ts
guidozoli: added

src/components/RadioGroup/RadioGroup.tsx
guidozoli: > > il linter si lamenta per la regola "react/no-array-index-key". Dovrei ignorarla per questa riga?
> 
> Nope. I do not suggest ignoring but, maybe, can use the key directly to `<Radio key={'${option.value}'} />` instead to a wrapper div?

Would this be ok also if value type is generic, so `option.value` can be an object?

src/components/RadioGroup/components/Radio/index.theme.ts
guidozoli: It is actually a design token. The figma was wrong. Now also for this value is used design token

src/components/RadioGroup/RadioGroup.tsx
Danielecina: yarn, the value can be an object. I suggest using indexing of the map

fredmaggiowski (dismissed):
LGTM

src/components/RadioGroup/components/Radio/index.theme.ts
```diff
@@ -0,0 +1,34 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ComponentsTheme } from '../../../ThemeProvider/Ant'
+import Theme from '../../../../themes/schema'
+
+/**
+ * Generates a Ant theme configuration for Radop component based on a theme configuration.
+ *
+ * @link https://ant.design/components/Radop#design-token
```
GiovannaMonti, line 25:
```suggestion
 * Generates a Ant theme configuration for Radio component based on a theme configuration.
 *
 * @link https://ant.design/components/radio#design-token
```

src/components/RadioGroup/components/Radio/index.theme.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param {Partial<Theme>} theme - theme configuration.
+ * @returns {Partial<ComponentsTheme>} The generated Input Ant theme configuration.
```
GiovannaMonti, line 28:
```suggestion
 * @returns {Partial<ComponentsTheme>} The generated Radio Ant theme configuration.
```

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -29,38 +29,35 @@ const defaults = {
   disabled: false,
 }
 
-export const RadioGroup = ({
+export const RadioGroup = <T, >({
```
fredmaggiowski, line 32: why a comma here?

```suggestion
export const RadioGroup = <T>({
```

guidozoli: otherwise the editor complains. It confuse it with an html element

fredmaggiowski: image

PR #718, opened 2024-11-20
fix(Button): Danger and Primary Outlined should have white background

malta895:
### Description

The Button when has hierarchy Danger or Primary, and type Outlined, should have a white background, and not transparent, as inherited from antd.

##### <Changed component name>
    - Button

### [IMPORTANT] PR Checklist

#### PR conventions

    > **NOTE**  

#### Additional code checks

### DIFF
--- a/src/components/Button/Button.module.css
+++ b/src/components/Button/Button.module.css
@@ -81,3 +81,11 @@
 .buttonText {
   padding: 0 var(--spacing-padding-xs, 4px);
 }
+
+.primaryOutlined {
+  background-color: var(--palette-action-secondary-main, #ffffff) !important;
+}
+
+.dangerOutlined {
+  background-color: var(--palette-action-secondary-main, #ffffff) !important;
+}
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -34,13 +34,15 @@ const {
   buttonGhost,
   buttonLink,
   buttonText,
+  primaryOutlined,
+  dangerOutlined,
 } = styles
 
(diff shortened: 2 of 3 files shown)


danielemarostica (approved):

src/components/Button/Button.module.css
```diff
@@ -81,3 +81,11 @@
 .buttonText {
   padding: 0 var(--spacing-padding-xs, 4px);
 }
+
+.primaryOutlined {
+  background-color: var(--palette-action-secondary-main, #ffffff) !important;
```
Danielecina, line 86: we can't use backgrounds like that. These are things that then cannot be themed

malta895: You mean we can't use backgrounds at all? There are also other usages in this CSS. 
How should I do it?

Danielecina: can u add a screenshot to show the problem?

malta895: image
image

As you can see in the first image the background is transparent, but it's supposed to be white.
The same applies to the Danger Outlined button

malta895: Ok I got your point. The problem is that our hierarchies, types and variants are not aligned 1:1 to antd ones, and our mapping has some issues, for example, we use ghost to obtain outline, which is wrong because it also make the background transparent.

The correct fix is to rewrite a complete mapping from our types to antd types, in a way that we replicate all the desired behaviors

malta895:
Closed as the issue is caused by a wrong prop mapping, explained in #721

PR #725, opened 2024-11-22
feat(Input): added custom names for input addon

nicmell:
### Description

The Input component now supports an object to be passed as a controlled value if the input has addons.
Additionally, you can specify custom prop names for the values in the addons by passing in their options a prop "name".
For the text input component, you can specify the prop valuePropName (defaults to "value") to customize the name of the value prop.

##### Input
    - The input now supports custom names to for the values of the input component and the addons

### [IMPORTANT] PR Checklist

#### PR conventions

    > **NOTE**  

#### Additional code checks

### DIFF
--- a/src/components/Input/Input.stories.tsx
+++ b/src/components/Input/Input.stories.tsx
@@ -283,8 +283,10 @@ export const WithCustomLogic = (): ReactElement => {
         }}
         isDisabled={inherited}
         value={value}
-        onChange={(_, {value}: any) => {
-            setValue(value)
+        onChange={(_, val: string | Record<string, unknown>) => {
+          if (typeof val === 'object') {
+            setValue(String(val.value))
+          }
         }}
       />
     </div>
--- a/src/components/Input/Input.stories.tsx
+++ b/src/components/Input/Input.stories.tsx
@@ -283,8 +283,8 @@ export const WithCustomLogic = (): ReactElement => {
         }}
         isDisabled={inherited}
         value={value}
-        onChange={(_, { value: val }) => {
-          setValue(val)
+        onChange={(_, {value}) => {
+            setValue(String(value))
         }}
       />
     </div>
--- a/src/components/Input/Input.stories.tsx
+++ b/src/components/Input/Input.stories.tsx
@@ -283,8 +283,8 @@ export const WithCustomLogic = (): ReactElement => {
         }}
         isDisabled={inherited}
         value={value}
-        onChange={(_, {value}) => {
-            setValue(String(value))
+        onChange={(_, {value}: any) => {
+            setValue(value)
         }}
       />
     </div>
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -35,18 +35,34 @@ import styles from './input.module.css'
 export const defaults = {
   ...baseInputDefaults,
   htmlType: Type.Text,
+  valuePropName: "value",
 }
 
 const DEFAULT_ICON_SIZE = 12 as never
 
 type AddonPosition = 'before' | 'after'
 
-const getAddonDefaultValue = (
-  position: AddonPosition,
-  props?: InputAddonProps
-): Record<string, unknown> => {
-  return props && (props.value !== undefined || props.defaultValue !== undefined)
-    ? { [position]: props.value || props.defaultValue } : {}
+const getAddonDefaultValue = (position: AddonPosition) =>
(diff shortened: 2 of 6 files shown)


src/components/Input/Input.tsx
```diff
@@ -35,18 +35,34 @@ import styles from './input.module.css'
...
+  valuePropName: string,
+  { before, after }: Record<AddonPosition, InputAddonProps | undefined>
+):
+  [
+    Record<string, unknown>,
+    React.Dispatch<React.SetStateAction<Record<string, unknown>>>
+  ] => {
+  return useState({
+    ...(typeof value === 'object' ? value : { [valuePropName]: value || '' }),
+    ...getAddonDefaultValue('before')(before),
```
Danielecina, line 63: Why function that returns another function and not directly `getAddonDefaultValue('before', before)`? I miss something?

src/components/Input/Input.tsx
```diff
@@ -77,47 +94,48 @@ export const Input = (
...
-      }
-      if (onChange) {
-        onChange(undefined, nextVal)
-      }
+      onChange(event, addonBeforeProp || addonAfterProp ? nextVal : nextValue)
     }
-
+  }, [addonAfterProp, addonBeforeProp, onChange, valuePropName, setVal, val])
+
+  const renderAddon = useCallback((
```
Danielecina, line 111: why return a component using `useCallback` method and not `useMemo`??

src/components/Input/Input.tsx
```diff
@@ -77,47 +94,48 @@ export const Input = (
...
+            onChange(undefined, nextVal)
+          }
+          if (onChangeAddon) {
+            onChangeAddon(nextValue)
+          }
+        }}
       />
     )
-  }, [isDisabled, onChange, val])
+  }, [isDisabled, onChange, setVal, val])
```
Danielecina, line 138: This renderAddon changes to every key down the user did for the addon input case. I think it loses a little bit of value in memoization

src/components/Input/Input.tsx
```diff
@@ -127,6 +145,18 @@ export const Input = (
     addonAfterProp && renderAddon('after', addonAfterProp)
   ), [addonAfterProp, renderAddon])
 
+  const getValue = useCallback((strOrObj?: string | Record<string, unknown>) => {
+    return typeof strOrObj === 'object' ? String(strOrObj[valuePropName]) : strOrObj
```
Danielecina, line 149: Is too much fragile -> 
Screenshot 2024-11-26 alle 10 19 10

src/components/Input/Input.stories.tsx
```diff
@@ -283,8 +283,10 @@ export const WithCustomLogic = (): ReactElement => {
         }}
         isDisabled={inherited}
         value={value}
-        onChange={(_, { value: val }) => {
-          setValue(val)
+        onChange={(_, val: string | Record<string, unknown>) => {
+          if (typeof val === 'object') {
+            setValue(String(val.value))
```
malta895, line 288: why is it `unknown` if then you convert it to a String?

src/components/Input/Input.tsx
nicmell: fixed

src/components/Input/Input.tsx
nicmell: this is just for silencing the linter, setVal is actually immutable as it's the setState method from a customHook

src/components/Input/Input.tsx
nicmell: we've added a more robust check with the function  isObject here https://github.com/mia-platform/design-system/blob/feat/input-addon-custom-value-names/src/utils/object.ts

src/components/Input/Input.stories.tsx
nicmell: This is an issue with typescript: the problem is the value parameter can be a string or an object, so our linter complains that "val.value" cannot be accessed in case "value" is a string.
It seems quite difficult to manage in typescript, so in the story the switch with the "typeof" cases is necessary, but it wouldn't if the project is in plain javascript, or even if you admit the "val" parameter to be of type "any" in the "onChange"  typing (our linter complains with using "any" in every context)

src/components/Input/Input.tsx
nicmell: correct. i've just fixed it

Danielecina (approved):

PR #729, opened 2024-11-26
feat(Alert): Added Alert component

nicmell:
### IDP-2129: Added alert component
Screenshot 2024-11-26 alle 09 22 43

IMPORTANT] PR Checklist

#### PR conventions

    > **NOTE**  

#### Additional code checks


src/components/Alert/Alert.module.css
```diff
@@ -0,0 +1,131 @@
...
+  border-left-width: 40px;
+  padding: var(--spacing-padding-lg, 16px);
+  gap: var(--spacing-gap-lg, 16px);
+  align-items: center;
+
+  background-color: var(--palette-purple-100, #EDE7FF);
+  border-color: var(--palette-purple-500, #986DF1);
+
+  & > svg {
+    position: absolute;
```
Danielecina, line 31: can be done in another way?

src/components/Alert/Alert.module.css
```diff
@@ -0,0 +1,131 @@
...
+}
+
+.description {
+  font-size: var(--typography-bodyS-fontSize, 14px);
+  line-height: var(--typography-bodyS-lineHeight, 16px);
+  font-weight: var(--typography-bodyS-fontWeight, 400);
+  color: var(--typography-bodyS-color, #636363);
+}
+
+.content {
```
Danielecina, line 96: can this be explained with a comment on this choice?

src/components/Alert/props.ts
```diff
@@ -0,0 +1,70 @@
...
+   */
+  description?: ReactNode
+
+  /**
+   * Custom children content to be rendered inside the alert.
+   */
+  children?: ReactNode
+
+  /**
+   * Indicates if the alert is closable.
```
malta895, line 42: I would specify the default here

src/components/Alert/Alert.module.css
nicmell: The problem i noticed is the typography component in the design-system uses a lot of !important in their classes, so it becomes hard to modify the style from the extern (it shouldn't be the case anyway, but for nested components can be a problem, ie. the CheckboxLabel > Checkbox > CheckboxGroup, where the style was the same of the BodyS component but the lineheight was different causing problems in spacings).

Additionally, using the css this way you don't have to write additional codes in the component, like:
if (isCompressed) {
  titleComponent = BodyS
} else {
  titleComponent = BodyM
}

nicmell: EDIT: i've seen the libary exposes also a component BodyX so the switch could be on the prop "size" instead, but using the css imo has always the benefit to limit the coding to the logic, and not to the styling

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,137 @@
...
+import MiExclamationCircle from '../../assets/icons/ExclamationCircle.svg'
+import { Type } from './types.ts'
+import XCircle from '../../assets/icons/XCircle.svg'
+import styles from './Alert.module.css'
+
+export const defaults = {
+  isClosable: true,
+}
+
+const getIcon = (type?: Type): IconComponent | undefined => {
```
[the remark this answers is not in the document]
nicmell: my bad. resolved

src/components/Alert/Alert.module.css
nicmell: @Danielecina i've just made some tries but seems quite hard. the problem is that the container is a flex and hence i'm not able to set the iconContainer to be the full height of the alert with "height" or any other property (except for some kind of absolute positioning)

src/components/Alert/props.ts
malta895: Ok, so even more so, I would explain it in the JSDoc, do you agree?

nicmell: done

Danielecina (approved):

PR #734, opened 2024-11-28
feat(Form): form implementation

nicmell:
### Component(Form): Form implementation

Form analysis implementation. 

A proof of concept of the implementation is included in this branch.
A description for the suggested implementation is provided in this branch at https://github.com/mia-platform/design-system/blob/IDP-94/form-analysis/src/components/Form/rfc.md.
Screenshot 2024-12-09 alle 11 41 15

### [IMPORTANT] PR Checklist

#### PR conventions

    > **NOTE**  

#### Additional code checks


src/components/Form/rfc.md
```diff
@@ -0,0 +1,441 @@
...
+        <Button onClick={form.resetFields}>
+          Clear form
+        </Button>
+        {defaultSubmitButton()}
+      </Flex>
+    )
+  }}
+/>
+```
+
```
Danielecina, line 189: We can Add the cases of 

```
<Form><button type=“submit”></Form>
<Form id=“myForm”></Form><button type=“submit” form=“myForm”>
```

src/components/Form/utils/validators.ts
```diff
@@ -0,0 +1,22 @@
...
+export type FormRule = AntFormRule
+
+export const required = (message?: string): FormRule => {
+  return {
+    message: message || 'The field is required',
+    required: true,
+  }
+}
+
+export const checkEquals = (fieldName: string, message?: string): FormRule => {
```
malta895, line 12: I would put a JSDoc over this, so that users of the library know how to use it well

src/components/Form/FormItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+    if (isValidElement(children)) {
+      return getDefaultFormItemProps(children)
+    }
+    return {}
+  }, [children])
+
+  const inputElement = useMemo(() => {
+    if (isValidElement(children)) {
+      return children
+    } else if (typeof children === 'function') {
```
malta895, line 87: the else if is redundant, the early return above would exit anyways

src/components/Form/FormItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+      }
+    case Switch:
+      return {
+        valuePropName: 'isChecked',
+      }
+    case RadioGroup:
+      return {
+        getValueFromEvent: ({ value }) => value,
+      }
+    default:
```
malta895, line 51: is this reachable? If we know all the possible types, I would error in case of unknown ones

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,66 @@
...
+  submitButtonProp: boolean | string | ((form: FormInstance) => ReactNode)
+): ReactNode => {
+  switch (typeof submitButtonProp) {
+  case 'boolean':
+    return submitButtonProp && defaultSubmitButton()
+  case 'string':
+    return defaultSubmitButton(submitButtonProp)
+  case 'function':
+    return submitButtonProp(form)
+  default:
```
malta895, line 33: We don 't accept other types, I would throw error

src/components/Form/Form.stories.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { TextArea } from '../TextArea'
+
+const meta = {
+  component: Form,
+  args: {},
+  argTypes: {},
+} satisfies Meta<typeof Form>
+
+export default meta
+// type Story = StoryObj<typeof meta>
```
malta895, line 27: is this to be removed?

src/components/Form/FormItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+  name?: string
+  label?: string
+  style?: CSSProperties
+  span?: number
+  isFullWidth?: boolean,
+  rules?: AntFormItemProps['rules']
+  valuePropName?: AntFormItemProps['valuePropName']
+  getValueFromEvent?: AntFormItemProps['getValueFromEvent']
+  children: ReactElement | ((props: RenderProps) => ReactNode)
+}
```
malta895, line 24: props are not documented. Could we create a dedicated types file with JSDocs?

malta895 (changes requested):

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+export const Form = <Values, >({
+  children,
+  style: styleProp,
+  submitButton: submitButtonProp = true,
+  columns = 2,
+  gap = 24,
```
fredmaggiowski, line 42: why accepting a style prop **and** a gap? should this default gap be ensured via the style default?

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,66 @@
...
+    columnGap: `${gap}px`,
+    ...styleProp,
+  }), [columns, gap, styleProp])
+
+  return (
+    <AntForm<Values> form={form} style={style} {...props}>
+      {children}
+      {
+        submitButtonProp && (
+          <div style={{ margin: `${gap}px 0`, gridColumn: `-1 / 1`, justifySelf: 'end' }}>
```
fredmaggiowski, line 59: shall we memoize this style?

src/components/Form/rfc.md
```diff
@@ -0,0 +1,467 @@
...
+### Layout Customization
+
+Given the limitations of configuring form fields via the `fields` prop, the proposal is to **deprecate this property** and instead define form items directly using React elements. The form items would be configured through the props supported by the component.
+
+This approach allows for a seamless coexistence of form items and custom elements within the DOM:
+
+```jsx
+<Form>
+  <FormItem name="input-name" {...fieldProps} />
```
fredmaggiowski, line 83: How will this be imported in the user codebase?

Do you expect to use

```js
import {Form, FormItem} from 'design-system'

function Comp({}){
  return <Form ... ><FormItem ...></Form>
}
```

of

```js
import {Form} from 'design-system'

function Comp({}){
  return <Form ... ><Form.Item ...></Form>
}
```

I'd prefer the last one (which is also similar to other subcomponent exports that have been done in the lib)

src/components/Form/Form.stories.tsx
```diff
@@ -0,0 +1,139 @@
...
+  return (
+    <div style={{ gridColumn: '-1 / 1', marginBottom: 24 }}>
+      <Card>
+        <pre>{JSON.stringify(values, null, '\t')}</pre>
+      </Card>
+    </div>
+  )
+}
+
+export const ComplexForm = (): ReactElement => {
```
fredmaggiowski, line 46: maybe a few stories about the submit button possibilities are useful for the final user

src/components/Form/utils/validators.ts
nicmell: sure! we are going to do this in the actual implementatation step

src/components/Form/FormItem.tsx
nicmell: correct

src/components/Form/FormItem.tsx
nicmell: yes, we will document the props in the next implementation step

src/components/Form/Form.tsx
nicmell: I would keep the prop style available in both the form and formItem mainly to override the default "display: grid" behaviour (even if not very likely). 
Additionally, the prop can be useful probably in future implementations for complex components like drawers with forms, modals with forms, etc.

src/components/Form/utils/validators.ts
nicmell: added jsdoc in https://github.com/mia-platform/design-system/blob/IDP-94/form-analysis/src/components/Form/Validators/index.ts

src/components/Form/FormItem.tsx
nicmell: props documented here https://github.com/mia-platform/design-system/blob/IDP-94/form-analysis/src/components/Form/Validators/index.ts

src/components/Form/Form.tsx
nicmell: done

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { PropsWithChildren, ReactElement, ReactNode, useMemo } from 'react'
+
+import { Button } from '../Button'
+
+type FormProps<Values> =
+  PropsWithChildren<AntFormProps<Values>>
+  & {
+  columns?: number
+  gap?: number
+  submitButton?: boolean | string | ((form: FormInstance) => ReactNode)
```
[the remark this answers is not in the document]
nicmell: Yes, it was temporary just for the proof of concept. Now all the props have been declared explicitly in the props.ts file

src/components/Form/Form.stories.tsx
nicmell: added few stories in here https://github.com/mia-platform/design-system/blob/IDP-94/form-analysis/src/components/Form/Form.stories.tsx#L133 and here for every form item https://github.com/mia-platform/design-system/blob/IDP-94/form-analysis/src/components/Form/FormItem/FormItem.stories.tsx

src/components/Input/Input.tsx
```diff
@@ -106,6 +107,7 @@ export const Input = (
     if (onChange) {
```
Danielecina, line 107: to remove the react-hooks/exhaustive-deps you can change the `setVal(nextVal)` in `setVal(currentValues => ({ ...currentValues, [valuePropName]: nextValue })`

nicmell: nextVal is needed the two lines below anyway:
```
      onChange(event, addonBeforeProp || addonAfterProp ? nextVal : nextValue)
```

src/components/Form/FormItem.tsx
nicmell: Yes, this is actually reachable: the element can have an unknown type if the input is implemented outside the library, so the __getDefaultFormItemProps__ should return an empty object (there are no known defaults to apply)

src/components/Form/Form.tsx
nicmell: This line is outdated, the new implementation is here, and the default is handled correctly:

https://github.com/mia-platform/design-system/pull/734/files#diff-7fa3d059e7802fedcc7a6d434f187ada28cb4aa1a92a95e0ccbf41935179ec26

malta895 (approved):

src/components/Input/Input.tsx
Danielecina: If you use the state update as it is now you run the risk of running into some bugs related to the asynchronicity of the useState https://react.dev/reference/react/useState#updating-state-based-on-the-previous-Ifate

You could send data to the onChange that is not yet updated.
If you use this approach, you can do everything within the callback function to guarantee the state update.

 ```
 setVal(curentObj => {
   const nextVal = { ...val, [valuePropName]: nextValue }
   if (onChange) {
     onChange(event, addonBeforeProp || addonAfterProp ? nextVal : nextValue)
   }
 })
 ```

Danielecina: I set this problem as an issue for the Input component

PR #776, opened 2024-12-18
fix(Input): fixed Ant props for Ant Design Input

### LINKED ISSUE #775: Input: Design System theme overrides Ant theme properties
Currently, when using an Ant Design `Input` component in a React application where both Ant Design and the Design System coexist, the Design System's generated theme overrides some properties of the Ant Design theme.

Although `Input` is used as an example here, this issue applies to any Design System component that wraps Ant Design components.

## The description of the bug or the rationale of your proposal
The `ThemeProvider` component wraps an `AntThemeProvider` component, which uses a function called `generateAntTheme` to generate a configuration for Ant Design based on an existing theme. This function allows for mapping component tokens to specific components.

For instance, in the `Input` component, the `paddingBlock` and `paddingInline` tokens are overridden and set to specific values to align with the Design System's `Input` component. While this works correctly when using only the Design System's `Input` component, it causes display issues when an Ant Design `Input` component is added to a page. Its padding values are overridden by the theme tokens, which affects the component's height and layout.

Below is a screenshot showing two Input components — one from the Design System and the other from Ant Design — illustrating how the `paddingBlock` and `paddingInline` properties impact the component height when overridden:

image

If the tokens are not set, the two components are correctly displayed:

image

In general, when creating a component that wraps Ant Design components, we should always check the effect of our theme on the Ant component.

Design System theme should not override Ant theme.

### Your environment
node:
v20.18.0
os:
Ubuntu 20.04.6 LTS

fedepini:
### Description

This PR has the goal of updating the Ant theme generated properties to make sure the `Input` component theme does not override the Ant `Input` component if used in the same App.

##### Input
    - updated them

### Addressed issue

Closes #775

### [IMPORTANT] PR Checklist

- [X] I am aware of standards and conventions adopted in this repository, defined in the [CONTRIBUTING.md file](https://github.com/mia-platform/design-system/blob/main/CONTRIBUTING.md)

#### PR conventions

- [X] PR title follows the `<type>(<scope>): <subject>` structure
- [X] The PR has been labeled according to the type of changes (e.g. enhancement, new component, bug).

    > **NOTE**  

#### Additional code checks

- [ ] Changes are covered by tests 
- [ ] Changes to components are accessible and documented in the Storybook
- [ ] The browser console does not contain errors

### DIFF
--- a/src/components/ThemeProvider/Ant.tsx
+++ b/src/components/ThemeProvider/Ant.tsx
@@ -18,6 +18,7 @@
 
 import { ConfigProvider, ThemeConfig, theme as baseAntTheme } from 'antd'
 import { ReactElement, useMemo } from 'react'
+import { omit } from 'lodash-es'
 
 import { DEFAULT_COLOR, DEFAULT_FONT_FAMILY, DEFAULT_FONT_SIZE } from './utils/themeDefaultStyle'
 import ButtonTheme from '../Button/Button.theme'
@@ -108,7 +109,7 @@ const generateAntTheme = ({ palette, typography, shape, spacing }: Partial<Theme
     Table: TableTheme({ palette, spacing }),
     Tree: TreeTheme({ palette, shape }),
     Typography: TypographyTheme({ typography }),
-    Input: InputTheme({ palette, shape, typography, spacing }),
+    Input: omit(InputTheme({ palette, shape, typography, spacing }), ['paddingBlock', 'paddingInline']),
     InputNumber: InputTheme({ palette, shape, typography, spacing }),
     Select: SelectTheme({ palette, shape, typography, spacing }),
     Tag: TagTheme({ palette }),


src/components/Input/Input.tsx
```diff
@@ -160,29 +172,31 @@ export const Input = (
...
-      maxLength={maxLength}
-      minLength={minLength}
-      placeholder={placeholder}
-      prefix={iconLeft && <Icon component={iconLeft} size={DEFAULT_ICON_SIZE} />}
-      suffix={iconRight && <Icon component={iconRight} size={DEFAULT_ICON_SIZE} />}
-      type={type}
-      value={inputValue}
-      onChange={handleChange}
-    />
+    <ConfigProvider theme={scopedAntTheme}>
```
Danielecina, line 175: it is not good to enter the provider here

nicmell: @Danielecina the pr relates this issue:
https://github.com/mia-platform/design-system/issues/775

I agree that the Provider should be used only as a wrapper over the application if possible.

PR #899, opened 2025-03-31
Added the possibility to change the body color

FrankieBortot:
### Description

Added the possibility to change body color into subtle and bold directly via dedicated prop.

### Checklist

- [ ] commit message and branch name follow conventions
- [ ] tests are included
- [ ] changes are accessible and documented from components stories
- [ ] you are not committing extraneous files or sensitive data
- [ ] necessary labels have been applied to this pull request (enhancement, bug, ecc.)

### DIFF
--- a/src/components/Typography/BodyX/BodyX.module.css
+++ b/src/components/Typography/BodyX/BodyX.module.css
@@ -38,10 +38,6 @@
   color: var(--palette-text-neutral-bold, #464646);
 }
 
-/* This class is empty because the main color is managed via Ant Theme */
-.main {
-}
-
 .subtle {
   color: var(--palette-text-neutral-subtle, #898989);
 }
--- a/src/components/Typography/BodyX/BodyX.module.css
+++ b/src/components/Typography/BodyX/BodyX.module.css
@@ -33,3 +33,15 @@
   font-size: var(--typography-bodyL-fontSize, 18px);
   line-height: var(--typography-bodyL-lineHeight, 24px);
 }
+
+.bold {
+  color: var(--palette-text-neutral-bold, #464646);
+}
+
+/* This class is empty because the main color is managed via Ant Theme */
+.main {
+}
+
+.subtle {
+  color: var(--palette-text-neutral-subtle, #898989);
+}
--- a/src/components/Typography/BodyX/BodyX.tsx
+++ b/src/components/Typography/BodyX/BodyX.tsx
@@ -26,7 +26,7 @@ import styles from './BodyX.module.css'
 
 const { Paragraph: AntParagraph } = AntTypography
 
-const { bodyS, bodyM, bodyL, bold, main, subtle } = styles
+const { bodyS, bodyM, bodyL, bold, subtle } = styles
 
 const { S, M, L } = Size
 
@@ -34,7 +34,6 @@ export const defaults = {
   copyable: false,
   ellipsis: false,
   isBold: false,
-  hierarchy: Hierarchy.Main,
 }
 
 /**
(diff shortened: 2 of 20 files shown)


src/components/Typography/BodyX/BodyX.module.css
```diff
@@ -33,3 +33,15 @@
   font-size: var(--typography-bodyL-fontSize, 18px);
   line-height: var(--typography-bodyL-lineHeight, 24px);
 }
+
+.bold {
+  color: var(--palette-text-neutral-bold, #464646);
+}
+
+/* This class is empty because the main color is managed via Ant Theme */
+.main {
```
Danielecina, line 42: if is not needed, I suggest to remove to avoid breaking everywhere a "bodyX" is used

src/components/Typography/BodyX/BodyX.tsx
```diff
@@ -46,18 +47,28 @@ export const BodyX = ({
...
+  ]), [hierarchy])
+  const bodyClassName = useMemo(() => classnames([
+    sizeClassName,
+    hierarchyClassName,
+  ]), [sizeClassName, hierarchyClassName])
 
   return (
     <AntParagraph
-      className={bodyClassName}
+      className= {bodyClassName}
```
fredmaggiowski, line 71: Please remove the space between `={`

Danielecina, line 71:
```suggestion
      className={bodyClassName}
```

src/components/Typography/BodyX/BodyX.tsx
```diff
@@ -46,18 +47,28 @@ export const BodyX = ({
...
 }: BodyXProps & BodyXSize): ReactElement => {
-  const bodyClassName = useMemo(() => classnames([
+  const sizeClassName = useMemo(() => classnames([
     size === S && bodyS,
     size === M && bodyM,
     size === L && bodyL,
   ]), [size])
+  const hierarchyClassName = useMemo(() => classnames([
+    hierarchy === Hierarchy.Bold && bold,
+    hierarchy === Hierarchy.Main && main,
```
Danielecina, line 61:
```suggestion
```

src/components/Typography/BodyX/BodyX.types.ts
```diff
@@ -33,3 +33,9 @@ export type BodyXSize = {
    */
   size: Size,
 }
+
+export enum Hierarchy {
+  Bold = 'bold',
+  Main = 'main',
```
Danielecina, line 39:
```suggestion
```

src/components/Typography/BodyX/BodyX.tsx
fredmaggiowski, line 61: I'd suggest not injecting any class if main is set

Since:

- applying this default changes tons of snapshots (we see some change here but it will also affect all of the snapshots for the projects that are using this component)
- the main class is currently empty

fredmaggiowski (changes requested):

src/components/Typography/BodyX/BodyX.module.css
fredmaggiowski: From a semantic standpoint it makes sense to have the hierarchy value imho 

I agree on removing this class since it affects too many snapshots

Danielecina: we can consider main as default and we don't need to distinguish by other type

src/components/Typography/BodyX/BodyX.tsx
```diff
@@ -20,20 +20,21 @@ import { ReactElement, useMemo } from 'react'
...
-const { bodyS, bodyM, bodyL } = styles
+const { bodyS, bodyM, bodyL, bold, main, subtle } = styles
 
 const { S, M, L } = Size
 
 export const defaults = {
   copyable: false,
   ellipsis: false,
   isBold: false,
+  hierarchy: Hierarchy.Main,
```
FrankieBortot, line 37: @fredmaggiowski @Danielecina should i also remove this default? (and also at row 50?)

Danielecina (approved):

fredmaggiowski (approved):

Issue #763, opened 2024-12-13
Virtualize select options


fredmaggiowski:
## The feature or bug you are proposing

I need to display thousands menu items in the select, each menu item needs to dynamically load a different icon. The current Select component does not let me implement this in a way that is productive without freezing the browser as soon as you click the select.

## A snippet of code for replicating the issue or showing the proposal usage if applicable

Attempting to see how the select behaves when using antd `optionRender`, I've prepared a story as such:

```ts
export const LotsOfOptions: Story = {
  args: {
    options: new Array(2000).fill('option')
      .map((val, i) => ({ label: `${val} ${i}`, value: i })),
    optionRender: (option) => {
      console.log('render', option)
      return <strong>{option.label}</strong>
    },
  },
}
```
however as soon as I open the select

<img width="1235" alt="immagine" src="https://github.com/user-attachments/assets/5be3b705-311a-418a-9bb5-903c3c6f3ce9" />

## The expected result for your bug

virtualization to happen and the render function to be invoked at need

davidebianchi:
The select by default supports the virtualization, but it is turned off with the `popupMatchSelectWidth` prop set to false.

From my point of view, we should give to the user the ability to change this behavior. By default I'd expect that the component virtualize, so we should breaking and change this prop.

PR #906, opened 2025-04-03
feat(Select): add optionRender props to select component

davidebianchi:
### Description

In this PR, I add the support to the `optionRender` to possibly add some information in the dropdown. For example, a description which can help to choose the correct option.

### [IMPORTANT] PR Checklist

#### PR conventions

    > **NOTE**  

#### Additional code checks

### DIFF
--- a/src/components/Select/Select.stories.tsx
+++ b/src/components/Select/Select.stories.tsx
@@ -19,6 +19,7 @@
 import { Meta, StoryObj } from '@storybook/react'
 import { FaDiamond } from 'react-icons/fa6'
 import { Flex } from 'antd'
+import { ReactNode } from 'react'
 
 import { Icon } from '../Icon'
 import { Select } from '.'
@@ -40,6 +41,7 @@ const options = [
   ...Array(5).keys(),
 ].map((id) => ({
   value: `value ${id + 1}`,
+  description: 'A description',
 }))
 
 export const Default: Story = {
@@ -138,5 +140,20 @@ export const MultipleDisabled: Story = {
     options,
     isMultiple: true,
     isDisabled: true,
+  },
+}
+
+export const CustomOptionRenderInDropdown: Story = {
+  args: {
+    defaultValue: options[0].value,
+    options,
+    optionRender: (option: {value?: ReactNode, description?: string}) => {
+      return (
+        <Flex align="center" gap={8}>
+          <Icon component={FaDiamond} size={16} />
+          <Typography.BodyM isBold>{option.value}</Typography.BodyM>
+        </Flex>
+      )
+    },
   },
 }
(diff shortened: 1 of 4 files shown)


nicmell:
Nice feature. I think a very desirable (necessary) feature would be also to have the possibility to render an option differently if the item is rendered as the selectedValue or an option (an example in the image below), but I'm not sure how this could be achieved.
In addition, it probably could be useful to have a default rendering for an optional description (inline as selected value, collapsed if as option) and an icon (not renderedas selected value, on the left as option), as also shown in the figure.

Screenshot 2025-04-03 alle 12 49 31

FrankieBortot:
Yes Nic, I would expect the behavior of the select to be exactly the same as the dropdown documented in our design system :)
So it should have a description, an optional icon, a "danger" state (for critical options), the "selected" state, and so on. If you check the "dropdown" component you can see all posibliities

Danielecina:
> Sì Nic, mi aspetterei che il comportamento del select fosse esattamente lo stesso del menu a discesa documentato nel nostro sistema di progettazione :) Quindi dovrebbe avere una descrizione, un'icona facoltativa, uno stato "pericoloso" (per le opzioni critiche), lo stato "selezionato" e così via. Se selezioni il componente "menu a discesa" puoi vedere tutte le possibilità

This is not a Select component. Is A Dropdown. The Two components work differently @FrankieBortot. 
If you expect the same experience we have problems because the Select component doesn't have this utility

Danielecina (approved):

davidebianchi:
It is quite simple to add also the feature you are thinking, it's another select option.

For the default option render, maybe we could create different sub components that can be used as option render easily? In this way, we could achieve both flexibility and a standard way to proceed. 

I think that this feature is ok to be merged, and in the future add the missing features. What do you think about?

nicmell (approved):

nicmell:
> It is quite simple to add also the feature you are thinking, it's another select option.
> 
> For the default option render, maybe we could create different sub components that can be used as option render easily? In this way, we could achieve both flexibility and a standard way to proceed.
> 
> I think that this feature is ok to be merged, and in the future add the missing features. What do you think about?

Sure, probably further implementations will provide a kind of default behaviour for this renderOption functionality.
@FrankieBortot we implemented the description, icon, etc in the Dropdown component but we hadn't much time to port it to the Select, and was questionable both the components should be refactored to use each other or not (the antd Select uses the Dropdown internally).

Meanwhile, I think this pr is fine and relates to this issue that I think can then be closed:
https://github.com/mia-platform/design-system/issues/763

davidebianchi:
No, it cannot be closed. That issue refers to the virtualization not the option render itself.

PR #938, opened 2025-05-06
feat(select): fix select virtualization, configuring popup width

### LINKED ISSUE #763: Virtualize select options

davidebianchi:
### Description

In this PR, resolve issues when the popup content is longer than the width.

Also, resolves #763 using select virtualization by default.

### Addressed issue

Resolves #763

### [IMPORTANT] PR Checklist

#### PR conventions

    > **NOTE**  

#### Additional code checks

- [ ] Changes are covered by tests 

### DIFF
--- a/src/components/Input/InputAddon.tsx
+++ b/src/components/Input/InputAddon.tsx
@@ -79,6 +79,7 @@ export const InputAddon = ({
         <Select
           isDisabled={isDisabled}
           isFullWidth={false}
+          popupMatchSelectWidth={false}
           onChange={onChange}
           {...props}
         />
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -83,6 +83,7 @@ export const Select = <ValueType, >(
     filterOption,
     optionFilterProp,
     dropdownRender,
+    popupMatchSelectWidth,
   }: SelectProps<ValueType>) : ReactElement => {
   const [open, setOpen] = useState(false)
 
@@ -122,7 +123,7 @@ export const Select = <ValueType, >(
       optionRender={optionRender}
       options={options}
       placeholder={placeholder}
-      popupMatchSelectWidth={false}
+      popupMatchSelectWidth={popupMatchSelectWidth}
       showSearch={Boolean(onSearch || filterOption)}
       suffixIcon={suffixIcon}
       tagRender={tagRender}
--- a/src/components/Select/props.ts
+++ b/src/components/Select/props.ts
@@ -135,4 +135,11 @@ export type SelectProps<ValueType = unknown> = BaseInputProps & {
    * Customize the rendering of the dropdown.
    */
   dropdownRender?: (menu: ReactNode) => ReactElement
+
+  /**
+   * Determine whether the popup menu and the select input are the same width.
+   * Default set min-width same as input. Will ignore when value less than select width.
+   * false will disable virtual scroll
+   */
+  popupMatchSelectWidth?: boolean | number
 }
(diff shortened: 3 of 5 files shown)


src/components/Input/InputAddon.tsx
```diff
@@ -79,6 +79,7 @@ export const InputAddon = ({
         <Select
           isDisabled={isDisabled}
           isFullWidth={false}
+          popupMatchSelectWidth={false}
```
Danielecina, line 82: can you add to the boolean props the is, has or can prefix?

davidebianchi: Is a bool or number props, not only boolean. Is it needed also in this cases the prefix? Could be maybe misleading?

davidebianchi: If needed, it is ok also should? So, `shouldPopupMatchSelectWidth` would be ok?

src/components/Select/props.ts
```diff
@@ -135,4 +135,11 @@ export type SelectProps<ValueType = unknown> = BaseInputProps & {
    * Customize the rendering of the dropdown.
    */
   dropdownRender?: (menu: ReactNode) => ReactElement
+
+  /**
+   * Determine whether the popup menu and the select input are the same width.
+   * Default set min-width same as input. Will ignore when value less than select width.
+   * false will disable virtual scroll
+   */
```
ThisIsDemetrio, line 143: I leave here a couple of suggestions on the jsdoc

```suggestion
  /**
   * Determine whether the popup menu and the select input should have the same width.
   * Default set `min-width` of the popup to match the input. 
   * This is ignored when the value is less than the select input's width.
   * Setting this to `false` will disable the virtual scroll
   */
```

src/components/Input/InputAddon.tsx
Danielecina: mmm so you can use a static width or a general true width. 
In this case I think the name may be fine as you propose. 
Are you sure it works? Because I remember the width is set by the css as “unset” to avoid automatic inheritance. if worts for me is approved

src/components/Select/Select.tsx
```diff
@@ -122,7 +123,7 @@ export const Select = <ValueType, >(
       optionRender={optionRender}
       options={options}
       placeholder={placeholder}
-      popupMatchSelectWidth={false}
+      popupMatchSelectWidth={popupMatchSelectWidth}
```
fredmaggiowski, line 126: is this a boolean prop? it's name should start with a verb such as `shouldPopupMatchSelectWidth` or `mustMatchPopupSelectWidth`

Danielecina (dismissed):

src/components/Select/Select.tsx
Danielecina: https://github.com/mia-platform/design-system/pull/938#discussion_r2077034626

fredmaggiowski (approved):
LGTM

